PR #50, opened 2019-02-22
 Expose accessibility configuration on view props

### LINKED ISSUE #39: Expose accessibility configuration on view props
Accessibility [labels, hints and traits](https://developer.apple.com/library/archive/documentation/UserExperience/Conceptual/iPhoneAccessibility/Making_Application_Accessible/Making_Application_Accessible.html#//apple_ref/doc/uid/TP40008785-CH102-SW5) should be available as Props on all components that render to views

hodovani:
Resolve #39

### DIFF
--- a/Sources/Tokamak/Components/Props/Style.swift
+++ b/Sources/Tokamak/Components/Props/Style.swift
@@ -30,6 +30,7 @@ public enum ContentMode {
 }
 
 public struct Style: Equatable {
+  public let accessibility: Accessibility?
   public let allowsEdgeAntialiasing: Bool?
   public let allowsGroupOpacity: Bool?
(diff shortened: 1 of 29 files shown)


Sources/Tokamak/Components/Props/Style.swift
```diff
@@ -5,7 +5,7 @@
 //  Created by Max Desiatov on 31/12/2018.
 //
 
-public protocol StyleProps {
+public protocol ViewProps {
```
MaxDesiatov: as `accessibility` is a property of `Style` after all, I think this should be renamed back to `StyleProps` to leave its use as clear as it was before.

PR #58, opened 2019-03-04
Add ScrollView

### LINKED ISSUE #32: ScrollView host component
Implementation of this component in `TokamakUIKit` would need special care to [make it work with auto layout](https://developer.apple.com/library/archive/documentation/UserExperience/Conceptual/AutolayoutPG/WorkingwithScrollViews.html). The approach I would try is to copy width constraint of a scroll view and its parent to the root subview. This would automatically do the same what people [need to do manually](https://stackoverflow.com/questions/48216808/programmatic-uiscrollview-with-autolayout) when working with programmatic `UIScrollView` and auto layout. We should also provide a flag in props to turn off these additional helper constraints if a user would like to avoid that. We also need to support both vertical and horizontal scrolling and inference of those additional constraints for both cases.

hodovani:
Resolves #32


MaxDesiatov:
run ci

MaxDesiatov:
trigger ci

PR #73, opened 2019-03-23
Fix CALayer-related props in TokamakAppKit

MaxDesiatov:
Currently none of the props that modify `CALayer` properties work in `TokamakAppKit`. The reason is that these properties [can only be update in `updateLayer` function of `NSView` subclass](https://developer.apple.com/documentation/appkit/nsview/1483580-updatelayer). This PR attempts to fix that.


MaxDesiatov:
Closing as outdated

PR #78, opened 2019-03-29
Update counter example with correct imports

twof:
`AnyNode` doesn't exist if `Tokamak` isn't imported.


MaxDesiatov:
Sorry I missed that, thank you very much @twof!

PR #79, opened 2019-03-29
Add TokamakLint module

hodovani:


Sources/TokamakLint/TokamakLint.swift
```diff
@@ -0,0 +1,115 @@
...
+        }
+      }
+    } catch {
+      print(error)
+    }
+  }
+
+  public func lintFile(_ path: String) {
+    do {
+      let fileURL: URL = URL(fileURLWithPath: path)
```
MaxDesiatov: same comment about explicit type annotation here

Tests/TokamakCLITests/SwiftSyntaxTests.swift
```diff
@@ -0,0 +1,35 @@
...
+    linter.lintFolder(path)
+  }
+
+  func testFile() {
+    let linter = TokamakLint()
+    let path = "/Users/hmi/Documents/maxDesiatov/Tokamak/Sources/Tokamak/Components/Host/Alert.swift"
+    linter.lintFile(path)
+  }
+
+//  func testPropsIsEquatable() {
```
MaxDesiatov: this code should be uncommented or removed before merge

Sources/TokamakLint/TokamakLint.swift
```diff
@@ -0,0 +1,115 @@
...
+          print("\(path):",
+                "\(structNode.range.startRow):",
+                "\(structNode.range.startColumn):",
+                " warning: Props is not Equatable")
+        }
+      }
+    } catch {
+      print(error)
+      return false
+    }
```
MaxDesiatov: these functions have a lot of similar structure with `do {} catch {}` but seem to handle the error in the same way. These functions could have `throws` annotation and the error could be handled uniformly higher in call stack and avoid repeated `do {} catch {}` structure

Sources/TokamakLint/TokamakLint.swift
```diff
@@ -0,0 +1,115 @@
...
+    } catch {
+      print(error)
+    }
+  }
+
+  public func isSwiftFile(_ path: String) -> Bool {
+    return path.contains(".swift")
+  }
+
+  public func hasTokamakImport(_ code: String) -> Bool {
```
MaxDesiatov: does this function have to be `public` and similarly functions above?

Sources/TokamakLint/TokamakLint.swift
```diff
@@ -0,0 +1,115 @@
...
+    return code.contains("import Tokamak")
```
MaxDesiatov: is there a more reliable way to check for `import` syntax nodes? This function would return `true` even for code that has `import Tokamak` commented out

Sources/TokamakLint/TokenVisitor.swift
```diff
@@ -0,0 +1,234 @@
...
+//
+//  Created by Matvii Hodovaniuk on 4/2/19.
+//
+
+import Foundation
+import SwiftSyntax
+
+class TokenVisitor: SyntaxVisitor {
+  public var tree = [Node]()
+  public var current: Node!
```
MaxDesiatov: do these properties have to be public? also, do we need to use an implicitly unwrapped optional for `current`?

Sources/TokamakLint/TokenVisitor.swift
```diff
@@ -0,0 +1,234 @@
...
+    if syntax.hasSuffix("Syntax") {
+      syntax = String(syntax.dropLast(6))
+    }
+
+    let node = Node(text: syntax)
+    node.range.startRow = row
+    node.range.startColumn = column
+    node.range.endRow = row
+    node.range.endColumn = column
+    if current == nil {
```
MaxDesiatov: when `current` is not an implicitly unwrapped optional, this could be rewritten with `if let current = current`

Sources/TokamakLint/TokenVisitor.swift
```diff
@@ -0,0 +1,234 @@
...
+      kind = String(kind.prefix(upTo: index))
+    }
+    if kind.hasSuffix("Keyword") {
+      kind = "keyword"
+    }
+
+    column += token.text.count
+  }
+
+  private func processTriviaPiece(_ piece: TriviaPiece) -> String {
```
MaxDesiatov: it would be great if this function had a doc comment that clarifies its purpose

Sources/TokamakLint/TokenVisitor.swift
```diff
@@ -0,0 +1,234 @@
...
+
+    token.leadingTrivia.forEach { piece in
+      let trivia = processTriviaPiece(piece)
+      current.token?.leadingTrivia += replaceSymbols(text: trivia)
+    }
+    processToken(token)
+    token.trailingTrivia.forEach { piece in
+      let trivia = processTriviaPiece(piece)
+      current.token?.trailingTrivia += replaceSymbols(text: trivia)
+    }
```
MaxDesiatov: I'm not sure why this trivia has to be processed for the checks, could you please add code comments that clarify the purpose of this in details?

Sources/TokamakLint/TokenVisitor.swift
```diff
@@ -0,0 +1,234 @@
...
+    case let .docBlockComment(text):
+      trivia += withSpanTag(class: "docBlockComment", text: text)
+      processComment(text: text)
+    default:
+      break
+    }
+    return trivia
+  }
+
+  private func withSpanTag(class c: String, text: String) -> String {
```
MaxDesiatov: same request for this function and similar functions below: their purpose is unclear, and there has to be comment that explains that in detailed

Sources/TokamakLint/TokenVisitor.swift
```diff
@@ -0,0 +1,234 @@
...
+    for node in typeNodes {
+      for type in node.children {
+        str.append("\(type.text)")
+      }
+    }
+    return str.contains(type)
+  }
+}
+
+class Node: Encodable {
```
MaxDesiatov: is this `Encodable` conformance used anywhere?

Tests/TokamakCLITests/SwiftSyntaxTests.swift
```diff
@@ -0,0 +1,35 @@
...
+//  Created by Matvii Hodovaniuk on 3/31/19.
+//
+
+import TokamakLint
+import XCTest
+
+final class SwiftSyntaxTests: XCTestCase {
+  func testAllFiles() {
+    let linter = TokamakLint()
+    let path = "/Users/hmi/Documents/maxDesiatov/Tokamak/"
```
MaxDesiatov: this path shouldn't be hardcoded

Tests/TokamakCLITests/SwiftSyntaxTests.swift
```diff
@@ -0,0 +1,35 @@
...
+    linter.lintFolder(path)
+  }
+
+  func testFile() {
+    let linter = TokamakLint()
+    let path = "/Users/hmi/Documents/maxDesiatov/Tokamak/Sources/Tokamak/Components/Host/Alert.swift"
```
MaxDesiatov: please avoid hardcoded paths

Tests/TokamakCLITests/TestPropsEquatable.swift
```diff
@@ -0,0 +1,16 @@
+//
+//  TestProps.swift
+//  Tokamak
+//
+//  Created by Matvii Hodovaniuk on 4/2/19.
+//
+
+import Foundation
+
+public protocol Default {
```
MaxDesiatov: is this file and its code used anywhere?

Tokamak.xcodeproj/xcshareddata/xcschemes/Tokamak-Package.xcscheme
```diff
@@ -0,0 +1,175 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
MaxDesiatov: I think this scheme can be deleted as unused

Tokamak.xcodeproj/xcshareddata/xcschemes/TokamakCLI.xcscheme
```diff
@@ -0,0 +1,88 @@
...
+            BuildableIdentifier = "primary"
+            BlueprintIdentifier = "Tokamak::TokamakCLI"
+            BuildableName = "TokamakCLI"
+            BlueprintName = "TokamakCLI"
+            ReferencedContainer = "container:Tokamak.xcodeproj">
+         </BuildableReference>
+      </BuildableProductRunnable>
+      <CommandLineArguments>
+         <CommandLineArgument
+            argument = "/Users/hmi/Documents/maxDesiatov/Tokamak/"
```
MaxDesiatov: this hardcode path has to be removed or replaced with a path calculated from variables

tokamak.dev/Linter.md
```diff
@@ -0,0 +1,24 @@
+## Tokamak Linter
+
+Tokamak Linter is module to check your Tokamak projects for known warning
+and show them in xcode or in terminal.
```
MaxDesiatov: "Xcode" starts with a capital "X"

tokamak.dev/Linter.md
```diff
@@ -0,0 +1,24 @@
...
+
+## How to run
+
+### Add to Xcode build phase
+
+- build TokamakCLI with Xcode
+- add path to TokamakCLI to build phase
```
MaxDesiatov: this should contain a snippet that could be added to a build phase, similar to what SwiftLint has in its README

Sources/TokamakLint/TokamakLint.swift
```diff
@@ -0,0 +1,106 @@
...
+    guard let enumerator = FileManager
+      .default
+      .enumerator(at: baseurl,
+                  includingPropertiesForKeys: resourceKeys,
+                  options: [.skipsHiddenFiles],
+                  errorHandler: { (url, error) -> Bool in
+                    print("directoryEnumerator error at \(url): ", error)
+                    return true
+      }) else {
+      fatalError("Enumeratrot is nil")
```
MaxDesiatov: please fix a typo here

Sources/TokamakLint/TokamakLint.swift
```diff
@@ -0,0 +1,106 @@
...
+    }
+    for case let fileURL as URL in enumerator {
+      if isSwiftFile(fileURL.path) {
+        print(fileURL.path)
```
MaxDesiatov: this `print` looks like a part of debug output and should be removed

Sources/TokamakLint/TokamakLint.swift
```diff
@@ -0,0 +1,106 @@
...
+  public func lintFile(_ path: String) throws {
+    let visitor = try walkParsedTree(path)
+    if hasTokamakImport(from: visitor) {
+      let structs = visitor.getNodes(get: "StructDecl", from: visitor.tree[0])
+      for structNode in structs {
+        let isInherited = visitor.isInherited(
+          node: structNode,
+          from: "Equatable"
+        )
+        print(isInherited)
```
MaxDesiatov: this `print` looks like a part of debug output

Sources/TokamakLint/TokamakLint.swift
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+
+  public func isPropsEquatable(_ path: String) throws -> Bool {
+    var res: Bool = false
+    let visitor = try walkParsedTree(path)
+    let structs = visitor.getNodes(get: "StructDecl", from: visitor.tree[0])
+    for structNode in structs where structNode.children[1].text == "Props" {
+      res = visitor.isInherited(node: structNode, from: "Equatable")
+      if !res {
```
MaxDesiatov: `if !` conditions with no `else` clauses are better expressed in Swift with `guard`, this could be rewritten as

```swift
guard res else { continue }
```

Sources/TokamakLint/TokamakLint.swift
```diff
@@ -0,0 +1,106 @@
...
+  ) {
+    print("\(path):", "\(row):", "\(column):", " \(type): \(message)")
+  }
+
+  private func isSwiftFile(_ path: String) -> Bool {
+    return path.contains(".swift")
+  }
+
+  private func hasTokamakImport(from visitor: TokenVisitor) -> Bool {
+    var isTokamakImportExist = false
```
MaxDesiatov: the name of this variable does not seem to be grammatically variable, a better alternative could be `doesTokamakImportExist`

Sources/TokamakLint/TokenVisitor.swift
MaxDesiatov: this comment is still valid: you can avoid using optional current in the `else` clause by using `if let current = current`

Sources/TokamakLint/TokenVisitor.swift
```diff
@@ -0,0 +1,158 @@
...
+    if current == nil {
+      tree.append(node)
+    } else {
+      current?.add(node: node)
+    }
+    current = node
+  }
+
+  override func visit(_ token: TokenSyntax) -> SyntaxVisitorContinueKind {
+    current?.text = escapeHtmlSpecialCharacters(token.text)
```
MaxDesiatov: would any of the code in this function execute when `current` is `nil`? If not, this could be optimised as `guard let current = current else { return }`

Tests/TokamakCLITests/TokamakLintTests.swift
```diff
@@ -0,0 +1,31 @@
...
+  func testFile() throws {
+    let linter = TokamakLint()
+    let path = "\(srcRoot)/Sources/Tokamak/Components/Host/Alert.swift"
+    try linter.lintFile(path)
+  }
+
+  func testPropsIsEquatable() throws {
+    let linter = TokamakLint()
+    let path = "\(srcRoot)/Tests/TokamakCLITests/TestPropsEquatable.swift"
+    XCTAssertTrue(try linter.isPropsEquatable(path))
```
MaxDesiatov: we should provide a negative test case here as well with a separate file, which does not have `Props` type conforming to `Equatable` and it would be verified that `isPropsEquatable` returns `false` for that file.

Tests/TokamakCLITests/TestPropsEquatable.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+// TokamakLintTests testPropsIsEquatable test use this file
+
+import Foundation
+
+public protocol Default {
+  var str: String { get }
+}
+
+struct Props: Equatable, Default {
```
MaxDesiatov: shouldn't the `Props` type here be a part of a component? The purpose of the check is to verify that only `Props` associated types within a component are `Equatable`

Sources/TokamakCLI/main.swift
```diff
@@ -0,0 +1,16 @@
...
+//
+
+import Foundation
+import TokamakLint
+
+do {
+  try lintFolder(CommandLine.arguments.first
+    ?? FileManager.default.currentDirectoryPath)
+} catch {
+  print("Can't lind folder")
```
MaxDesiatov: please also log the actual error here

Sources/TokamakLint/TokenVisitor.swift
```diff
@@ -0,0 +1,159 @@
...
+      kind = String(kind.prefix(upTo: index))
+    }
+    if kind.hasSuffix("Keyword") {
+      kind = "keyword"
+    }
+
+    column += token.text.count
+  }
+
+  private func escapeHtmlSpecialCharacters(_ string: String) -> String {
```
MaxDesiatov: please add a code comment here explaining why this function is needed and how it's used

Sources/TokamakLint/TokamakLint.swift
```diff
@@ -0,0 +1,105 @@
...
+
+private func walkParsedTree(_ path: String) throws -> TokenVisitor {
+  let fileURL = URL(fileURLWithPath: path)
+  let parsedTree = try SyntaxTreeParser.parse(fileURL)
+  let visitor = TokenVisitor()
+  parsedTree.walk(visitor)
+  return visitor
+}
+
+private func formatError(
```
MaxDesiatov: please add a unit-test for `formatError` function

PR #89, opened 2019-05-03
Add SwiftCLI dependency to TokamakCLI

hodovani:
Add CLI to TokamakCLI to lint file or folder with TokamakLint.

### DIFF
--- a/Sources/TokamakCLI/main.swift
+++ b/Sources/TokamakCLI/main.swift
@@ -6,27 +6,27 @@
 //
 
 import Foundation
-import TokamakLint
 import SwiftCLI
+import TokamakLint
 
 class LintCommand: Command {
-    let name = "lint"
-    let shortDescription = "Lint folder or file"
-    let path = Parameter()
+  let name = "lint"
+  let shortDescription = "Lint folder or file"
+  let path = Parameter()
 
-    func execute() throws {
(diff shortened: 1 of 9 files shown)


Sources/TokamakCLI/main.swift
```diff
@@ -7,11 +7,29 @@
...
+    func execute() throws {
+        do {
+          let srcRoot = ProcessInfo.processInfo.environment["SRCROOT"]!
+            if path.value.contains(".swift") {
+                try lintFile("\(srcRoot)/\(path.value)")
+            } else {
+                try lintFolder("\(srcRoot)/\(path.value)")
+            }
+        } catch {
+          print("Can't lint folder")
```
MaxDesiatov: this error text won't be applicable when user passed a file instead of a folder to lint

Sources/TokamakCLI/main.swift
```diff
@@ -7,11 +7,29 @@
...
+          print(error)
+        }
+    }
 }
+
+let TokamakCLI = CLI(name: "TokamakCLI", version: "1.0.0", description: "Tokamka CLI tools")
```
MaxDesiatov: please set the version to correspond to the current version of the project: 0.1.2

PR #100, opened 2019-05-28
Fix one render function rule 

hodovani:
One render function rule checks all components, not only hooked components.


Sources/TokamakLint/Helpers/GetComponents.swift
```diff
@@ -1,27 +1,30 @@
...
 //  TokamakLint
 //
-//  Created by Matvii Hodovaniuk on 5/23/19.
+//  Created by Matvii Hodovaniuk on 5/28/19.
 //
 
 import SwiftSyntax
 
+let hookedComponentProtocols = ["CompositeComponent", "LeafComponent"]
+let componentsProtocols = ["CompositeComponent", "LeafComponent",
```
MaxDesiatov: please rename this to `componentProtocols` for consistency, you can also reuse `hookedComponentProtocols` here via array concatenation

Sources/TokamakLint/Rules/RenderGetsHooksRule.swift
```diff
@@ -15,38 +15,49 @@ struct RenderGetsHooksRule: Rule {
...
-        with: SyntaxKind.simpleTypeIdentifier.rawValue
-      ).filter {
-        guard let children = $0.children.first else { return false }
-        return children.text == "Hooks"
-      }
+        let hooksArgument = functionSignature.children(
+          with: SyntaxKind.simpleTypeIdentifier.rawValue
+        ).filter {
+          guard let children = $0.children.first else { return false }
+          return children.text == "Hooks"
```
MaxDesiatov: this can be simplified as `{ $0.children.first == "Hooks" }`

Sources/TokamakLint/Rules/RenderGetsHooksRule.swift
```diff
@@ -15,38 +15,49 @@ struct RenderGetsHooksRule: Rule {
...
-          line: renderFunction.range.startRow,
-          character: renderFunction.range.startColumn
-        )
-      )] }
-      guard let functionSignature = codeBlock.firstChild(
-        of: SyntaxKind.functionSignature.rawValue
-      ) else { return [] }
+      // search for render function
+      let structs = visitor.root.components(hookedComponentProtocols)
+      guard !structs.isEmpty else { return [] }
```
MaxDesiatov: this check is redundant as the following `forEach` won't execute anything anyway if `structs` is empty

Sources/TokamakLint/Rules/RenderGetsHooksRule.swift
```diff
@@ -15,38 +15,49 @@ struct RenderGetsHooksRule: Rule {
...
-          ruleDescription: OneRenderFunctionRule.description,
-          location: Location(
-            file: visitor.path,
-            line: renderFunction.range.startRow,
-            character: renderFunction.range.startColumn
-          )
-        )]
+        guard !hooksArgument.isEmpty else {
+          violations.append(StyleViolation(
+            ruleDescription: OneRenderFunctionRule.description,
```
MaxDesiatov: shouldn't this use a different description?

PR #104, opened 2019-06-01
Add TokamakLogger

hodovani:


Sources/TokamakLint/Reporters/TokamakLogger.swift
```diff
@@ -0,0 +1,132 @@
...
+  private var _logLevel: Logger.Level = .info
+
+  // metadata storage
+  public var metadata: Logger.Metadata = [:]
+
+  public init(label: String) {}
+
+  public var outputs = [Outputs.file]
+
+    public var _path: URL?
```
MaxDesiatov: formatting is off

Sources/TokamakLint/Reporters/TokamakLogger.swift
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  public let rawValue: Int
+
+  static let stdout = Outputs(rawValue: 1)
+  static let file = Outputs(rawValue: 2)
+}
+
+public struct TokamakLogger: LogHandler {
+  private static var overrideLogLevel: Logger.Level?
```
MaxDesiatov: this variable is not used anywhere

Sources/TokamakLint/Reporters/TokamakLogger.swift
```diff
@@ -0,0 +1,132 @@
...
+      }
+      return _path
+    }
+
+    set {
+      _path = newValue
+    }
+  }
+
+  public var logLevel: Logger.Level {
```
MaxDesiatov: `_logLevel` isn't used anywhere other than this computed variable. It would be much simpler if `logLevel` was a simple stored property without proxying to `_logLevel`.

Sources/TokamakLint/Reporters/TokamakLogger.swift
```diff
@@ -0,0 +1,132 @@
...
+//  TokamakLogger.swift
+//  TokamakLint
+//
+//  Created by Matvii Hodovaniuk on 5/30/19.
+//
+
+import Foundation
+import Logging
+
+enum Output {
```
MaxDesiatov: this enum isn't used anywhere

Tests/TokamakCLITests/TokamakLint/TokamakLoggerTest.swift
```diff
@@ -0,0 +1,27 @@
...
+//
+//  Created by Matvii Hodovaniuk on 5/31/19.
+//
+
+import Logging
+@testable import TokamakLint
+import XCTest
+
+final class TokamakLoggerTests: XCTestCase {
+  func testLogToFile() throws {
```
MaxDesiatov: there are no assertions in this test, can you please clarify under what conditions this test would fail?

Sources/TokamakLint/Reporters/TokamakLogger.swift
```diff
@@ -0,0 +1,132 @@
...
+      level: level,
+      message: message,
+      metadata: .none,
+      file: "",
+      function: "",
+      line: 0
+    )
+  }
+
+  func write(_ string: String) {
```
MaxDesiatov: can this function be `private`?

Sources/TokamakLint/Reporters/TokamakLogger.swift
```diff
@@ -0,0 +1,132 @@
...
+    let fm = FileManager.default
+    if !fm.fileExists(atPath: path.absoluteString) {
+      fm.createFile(
+        atPath: path.absoluteString,
+        contents: data, attributes: nil
+      )
+    } else {
+      var file = FileHandle(forReadingAtPath: path.absoluteString)
+      _ = String(describing: file?.readDataToEndOfFile())
+      file = FileHandle(forWritingAtPath: path.absoluteString)
```
MaxDesiatov: it's unclear what this code is doing and why the filehandle is opened twice for every log message. I think it would more effective to open `FileHandle` once in the initializer and to reuse the handle across writes

Sources/TokamakLint/Reporters/TokamakLogHandler.swift
```diff
@@ -0,0 +1,101 @@
...
+//  TokamakLint
+//
+//  Created by Matvii Hodovaniuk on 5/30/19.
+//
+
+import Foundation
+import Logging
+
+public struct LogFileCreationError: Error {
+  public init() {}
```
MaxDesiatov: this can be removed now as `LogFileCreationError` is no longer created outside of `TokamakLint` module

Sources/TokamakLint/Reporters/TokamakLogHandler.swift
```diff
@@ -0,0 +1,101 @@
...
+}
+
+struct StringConversionToDataError: Error {}
```
MaxDesiatov: In fact, both of these errors could be merged now:

```swift
enum TokamakLintError: Error {
  case logFileCreationFailed
  case logMessageEncodingFailed
}
```

Sources/TokamakLint/Reporters/TokamakLogHandler.swift
```diff
@@ -0,0 +1,101 @@
...
+//  TokamakLogHandler.swift
+//  TokamakLint
+//
+//  Created by Matvii Hodovaniuk on 5/30/19.
+//
+
+import Foundation
+import Logging
+
+enum TokamakLintError: Error {
```
MaxDesiatov: please make it public, so that end users are able to catch this error

PR #116, opened 2020-06-28
Add TextField, SecureField, and a progress document

j-f1:
Currently, it seems that changing the `text` binding doesn’t actually affect the rendered input. I’m not sure why this happens, but I’m guessing it’s because the `value` attribute isn’t being set as `<input>.value`.

I’ve added two TextFields to the demo, which are supposed to have their values synced. Pressing enter in the second one will increment the commit counter.


j-f1:
Looks like Danger is failing at least because it can’t access the GitHub token since this PR comes from a fork.

MaxDesiatov:
Thanks for the PR, I'll have a look a bit later today! `danger-lint` failing is a known issue, I'll turn it off in `main` so that this doesn't confuse anyone anymore.

progress.md
```diff
@@ -0,0 +1,137 @@
+)# Progress
```
MaxDesiatov, line 1: This is a file I should've created from the start, thank you for doing this! Could you also create a `docs` directory and move it there? Also, seems that `)` in the beginning is not needed:

```suggestion
# Progress
```

j-f1: How detailed should this be? I originally had it in a list format which could’ve displayed individual method names but I figured the table would give people a good overview of what is usable and where they can jump in.

MaxDesiatov: Oh, what is currently there already looks great. Maybe I'd tweak table headers, they look kind of weird as empty thin rows, but don't sweat it. I think the API comparison should be automated at some point, but for the list of basic fundamental views this should be more than enough from my perspective, unless you think otherwise.

Hey @carson-katri, I'd be happy to hear your opinion on this, what do you think?

MaxDesiatov:
I think the text binding issues may be caused by bugs in the `State`/`Binding` code, investigating now...

MaxDesiatov:
Turns out it's not a state/binding issue, but a more prosaic oversight that's fixed in #117.

j-f1:
Is there a way to hook into the `update` function? I need to update the actual `value` property of the input element, not the `value` attribute.

MaxDesiatov:
Is there a signifcant difference? I've tested your branch locally with #117 applied on top of it and it worked, or am I missing something?

j-f1:
If you type in both fields they will stop updating.

MaxDesiatov:
huh, good point, on it...

progress.md
carson-katri: I can explore an automated approach. In the meantime, you could do the table with HTML to get rid of the header if you wanted to fix them.

MaxDesiatov: I don't think headers are that important, so wouldn't mind if they are kept as is in this PR.

MaxDesiatov:
I updated #117 to update properties, not attributes, so should work now when combined with this PR.

j-f1:
Is it intentional that `VStack` can have a max of 10 children?

j-f1:
Updated. Note that the newly added text field is still editable despite that not being the intention.

carson-katri:
@j-f1 This is a limitation of the `ViewBuilder`, and is present in SwiftUI as well. `ViewBuilder` takes in arguments and builds a `TupleView` which is generated to have up to 10 children. If you use `ForEach` or a `Group` you can go past that limit.

j-f1:
OK cool I reviewed native SwiftUI and the text field’s behavior is the same there 🤷

j-f1:
I tried to implement [the other TextField initializer](https://developer.apple.com/documentation/swiftui/textfield/init(_:value:formatter:oneditingchanged:oncommit:)-69262) but it looks like `Foundation` (and therefore the required `Formatter`) are not yet available?

j-f1:
I dislike the way the SecureField is largely a copy-paste of TextField but I can’t think of a better way since you can’t extend structs.

carson-katri:
Could you do a protocol that has all the fields and a `isSecure` flag?

MaxDesiatov:
`Foundation` is available if you build the package in a special way, but I'd like to avoid it as much as possible. It's a huge dependency, and I'm afraid that even when SwiftWasm has dead code elimination working properly, I'm not sure we'll be able to strip out unused bits of Foundation cleanly, especially due to how big it is and how it also proxies a lot of stuff to CoreFoundation. On the other hand, without Foundation we won't get the `Date` type, which would be frustrating.

We need to tweak `carton` so that it links Foundation automatically, but I think for now, please leave out the bits that require Foundation, and maybe add a comment about what's missing.

j-f1:
> Could you do a protocol that has all the fields and a `isSecure` flag?

I gave that a try and it didn’t cut down on the duplication a ton.

Sources/TokamakCore/Views/TextField.swift
```diff
@@ -0,0 +1,53 @@
...
+    onEditingChanged: @escaping (Bool) -> () = { _ in },
+    onCommit: @escaping () -> () = {}
+  ) where S: StringProtocol {
+    label = Text(title)
+    textBinding = text.projectedValue
+    editingChangedAction = onEditingChanged
+    commitAction = onCommit
+  }
+
+  // Currently missing feature: use a Formatter to control the value of the TextField
```
MaxDesiatov, line 43: **suggestion:** I think it would make sense to highlight this with a `FIXME`.

```suggestion
  // FIXME:
  // Currently missing feature: use a Formatter to control the value of the TextField
```

carson-katri:
SwiftUI doesn't use a protocol, and if it's good enough for Apple then it's good enough for me 😄

PR #117, opened 2020-06-28
Update DOM node properties and listeners on renderer update

MaxDesiatov:
This fixes an issue discovered in #116 by updating DOM attributes when `AnyHTML.update(dom:)` is called from the renderer update function. I'm still looking for a good solution for reinstalling all event listeners correctly in the same function, all suggestions welcome.

### DIFF
--- a/Sources/TokamakDOM/Views/HTML.swift
+++ b/Sources/TokamakDOM/Views/HTML.swift
@@ -91,7 +91,7 @@ extension HTML where Content == EmptyView {
     _ attributes: [String: String] = [:],
     listeners: [String: Listener] = [:]
   ) {
-    self = HTML(tag, attributes) { EmptyView() }
+    self = HTML(tag, attributes, listeners: listeners) { EmptyView() }
   }
 }
 
--- a/Sources/TokamakDOM/Views/HTML.swift
+++ b/Sources/TokamakDOM/Views/HTML.swift
@@ -42,10 +42,21 @@ extension AnyHTML {
     """
   }
 
-  func update(dom: JSObjectRef) {
-    // FIXME: handle attributes and listeners here
+  func update(dom: DOMNode) {
+    // FIXME: is there a sensible way to diff attributes and listeners to avoid
+    // crossing the JavaScript bridge and touching DOM if not needed?
+
+    // @carson-katri: For diffing, could you build a Set from the keys and values of the dictionary,
+    // then use the standard lib to get the difference?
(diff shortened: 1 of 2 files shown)


carson-katri:
For diffing, could you build a `Set` from the keys and values of the dictionary, then use the standard lib to get the difference?

For diffing the Listeners `ObjectIdentifier` may work.

MaxDesiatov:
Great shout about the stdlib diffing functions, I'm going to add that as a comment. `ObjectIdentifier` will not work for closures though, it only works for class instances as far as I'm ware.

carson-katri:
Could you compare JSFunctions?

MaxDesiatov:
Yes, but our users don't create those directly, i.e. `Button("Increment") { count += 1 }` written by the user will create a new `JSFunction` each time, because there's no way to distinguish whether the closure they supplied is different from the existing one.

MaxDesiatov:
And even if they created those manually, `Button("Increment", action: JSClosure { count += 1 })` would create a new `JSClosure` on every renderer update anyway.

Sources/TokamakDOM/Views/HTML.swift
```diff
@@ -42,10 +42,21 @@ extension AnyHTML {
...
   }
 
-  func update(dom: JSObjectRef) {
-    // FIXME: handle attributes and listeners here
+  func update(dom: DOMNode) {
+    // FIXME: is there a sensible way to diff attributes and listeners to avoid
+    // crossing the JavaScript bridge and touching DOM if not needed?
+
+    // @carson-katri: For diffing, could you build a Set from the keys and values of the dictionary,
+    // then use the standard lib to get the difference?
```
carson-katri, line 50: To do this I'd suggest `zip`, but tuples aren't `Hashable`, so maybe something like:
```swift
fileprivate struct KVP : Hashable {
    let key: String
    let value: String
}
...
Set(attributes.map { KVP(key: $0.0, value: $0.1) })
.subtracting(...)
```

MaxDesiatov: Totally, I just added that as a comment so that diffing is handled as a separate PR. Maybe even in the future when we have some kind of performance benchmarks so that we can carefully tweak for performance.

MaxDesiatov:
Ready for review now 🙂

carson-katri:
Everything seems to be working/looking good 👍

PR #119, opened 2020-06-29
Shape, Path, and some related modifiers

carson-katri:
This implements SwiftUIs `Shape`, `ShapeStyle`, etc. along with `background` and `border` modifiers, and the `CGAffineTransform` matrix.

The order of modifiers matters, as in SwiftUI, and the ordering should match SwiftUIs closely.

Here the background hugs the text, and the border is offset by the padding:
```swift
Text("Hello, world!")
  .background(Color.red)
  .padding()
  .border(Color.blue)
```

Here the background and border are offset by the padding:
```swift
Text("Hello, world!")
  .padding()
  .background(Color.red)
  .border(Color.blue)
```

`Color` conforms to `ShapeStyle` as well as `View` to match SwiftUI.

You can also use `Shapes` (`Rectangle`, `Ellipse`, `Capsule`, etc.) as SVG. However, this is not fully implemented in the renderer yet. This is a very large task and could probably be split up across multiple PRs for different features of `Shape`. Currently the SVG container is sized at 0 unless you explicitly specify the `CGSize`.


Sources/TokamakCore/CGStubs.swift
```diff
@@ -16,3 +16,194 @@
...
+    self.ty = ty
+  }
+
+  /// Returns an affine transformation matrix constructed from a rotation value you provide.
+  /// - Parameters:
+  ///   - angle: The angle, in radians, by which this matrix rotates the coordinate system axes.
+  ///            A positive value specifies clockwise rotation and a negative value specifies counterclockwise rotation.
+  public init(rotationAngle angle: CGFloat) {
+    // FIXME: How do we use sin/cos in Wasm?
+    //        With Darwin: self.init(a: cos(angle), b: sin(angle), c: -sin(angle), d: cos(angle), tx: 0, ty: 0)
```
j-f1, line 113: > WebAssembly doesn’t include its own math functions like sin, cos, exp, pow, and so on. WebAssembly’s strategy for such functions is to allow them to be implemented as library routines in WebAssembly itself (note that x86’s sin and cos instructions are slow and imprecise and are generally avoided these days anyway). Users wishing to use faster and less precise math functions on WebAssembly can simply select a math library implementation which does so.

([WASM FAQ](https://webassembly.org/docs/faq/))

j-f1: Maybe best off hand-coding it? See [Wikipedia](https://en.wikipedia.org/wiki/Lookup_table#Computing_sines) for one way based on a Taylor series.

Sources/TokamakCore/Shapes/Path.swift
```diff
@@ -0,0 +1,291 @@
...
+    storage = .rect(rect)
+  }
+
+  public init(roundedRect rect: CGRect,
+              cornerSize: CGSize,
+              style: RoundedCornerStyle = .circular) {
+    storage = .roundedRect(FixedRoundedRect(rect: rect,
+                                            cornerRadius: cornerSize.width,
+                                            style: style))
+    // FIXME: This currently doesn't support an actualy CGSize. I'm not sure how that works or what it does.
```
j-f1, line 75: Looks like if you do `.clipShape(RoundedRectangle(cornerSize: CGSize(width: 20, height: 10), style: .circular))` you get this:
Screenshot_2020-06-29 12 33 11

If you use `.continuous`, it looks like it just uses the `width` value.

Sources/TokamakCore/Shapes/Path.swift
```diff
@@ -0,0 +1,291 @@
...
+
+  public func strokedPath(_ style: StrokeStyle) -> Self {
+    Self(storage: .stroked(StrokedPath(path: self, style: style)))
+  }
+
+  public func trimmedPath(from: CGFloat, to: CGFloat) -> Self {
+    Self(storage: .trimmed(TrimmedPath(path: self, from: from, to: to)))
+  }
+
+  //  In SwiftUI, but we don't have CGPath...
```
j-f1, line 136:
```suggestion
  //  FIXME: In SwiftUI, but we don't have CGPath...
```

Sources/TokamakDOM/Shapes/Path.swift
```diff
@@ -0,0 +1,83 @@
...
+      return AnyView(EmptyView())
+    case let .rect(rect):
+      return AnyView(AnyView(HTML("rect", ["width": "\(max(0, rect.size.width))",
+                                           "height": "\(max(0, rect.size.height))"]
+          .merging(stroke, uniquingKeysWith: uniqueKeys))))
+    case .ellipse:
+      return AnyView(HTML("ellipse", ["cx": "50%", "cy": "50%", "rx": "50%", "ry": "50%"]
+          .merging(stroke, uniquingKeysWith: uniqueKeys)))
+    case let .roundedRect(roundedRect):
+      return AnyView(HTML("rect", [
```
j-f1, line 46: Can you add a `FIXME` to support the `.continuous` style?

Sources/TokamakDOM/Shapes/Shape.swift
```diff
@@ -0,0 +1,43 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//  Created by Carson Katri on 6/29/20.
+//
+
+import TokamakCore
+
+// This isn't identical to the SwiftUI API, but is a simple overload to support HTML `border`.
+// fileprivate struct _BorderModifier : ViewModifier, DOMViewModifier {
```
j-f1, line 21: Why is this commented?

Sources/TokamakDOM/Shapes/_ShapeView.swift
```diff
@@ -0,0 +1,35 @@
...
+//  Created by Carson Katri on 6/29/20.
+//
+
+import TokamakCore
+
+extension _ShapeView: ViewDeferredToRenderer {
+  public var deferredBody: AnyView {
+    let path = shape.path(in: .zero).deferredBody
+    if let fillColor = style as? Color {
+      return AnyView(HTML("div", ["style": "fill: \(fillColor.description)"]) { path })
```
j-f1, line 24: Does `fill` work in HTML?

Sources/TokamakCore/CGStubs.swift
MaxDesiatov: These are available in [WASI](wasi.dev), which you currently can get with `import Glibc`. I do plan to rename that module to `WASI`, so you will have to write `import WASI` in the future, but for now `import Glibc` should work.

Sources/TokamakDOM/Shapes/Shape.swift
carson-katri: Ah, sorry. I found a better way that does match the SwiftUI API, so I commented that in case it didn't work, but forgot to delete it.

Sources/TokamakDOM/Shapes/_ShapeView.swift
carson-katri: It applied it to the child `svg` tags.

Sources/TokamakCore/Shapes/Path.swift
carson-katri: Thanks for the reference! I implemented it to match.

Sources/TokamakCore/Shapes/Path.swift
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  public init(_ callback: (inout Self) -> ()) {
+    var base = Self()
+    callback(&base)
+    self = base
+  }
+
+  public init?(_ string: String) {
+    // FIXME: Somehow make this from a string?
```
j-f1, line 96: Here’s what `Path(roundedRect: CGRect(x: 0, y: 0, width: 50, height: 100), cornerRadius: 5, style: .circular)` looks like:

```
"50 50 m 50 95 l 50 97.7614 47.7614 100 45 100 c 5 100 l 2.23858 100 3.38177e-16 97.7614 0 95 c 5.55112e-15 5 l 5.88929e-15 2.23858 2.23858 -1.69088e-16 5 0 c 45 0 l 47.7614 1.69088e-16 50 2.23858 50 5 c h"
```

and here’s `Path(ellipseIn: CGRect(origin: .zero, size: CGSize(width: 100, height: 100)))`:

```
"100 50 m 100 77.6142 77.6142 100 50 100 c 22.3858 100 0 77.6142 0 50 c 0 22.3858 22.3858 0 50 0 c 77.6142 0 100 22.3858 100 50 c h"
```

not exactly sure what it means 🤷

carson-katri: Looks like [path commands](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/d)!

j-f1: Sorta but they have all the commands lowercased and it’s a postfix notation

j-f1: Looks to be similar to but different from what you get when you do `String(describing: CGPath(...))` so this can probably just be skipped.

carson-katri: You mean that printing the `CGPath` and printing the `Path` are the same? Because I would still need to implement it for `CGPath`...

j-f1: They’re similar but different.

j-f1: Here’s what I’ve got so far (in JS, using PEG.js for parsing because it’s something I’ve used before): https://observablehq.com/d/c4d2be55a2e0b70a

```js
Path = first: Command rest: (s cmd: Command { return cmd })* { return [first, ...rest] }

Command = Move / Line / Curve / Home

Move "move"
  = x: number s y: number s "m"
    { return { command: 'moveto', x, y } }
Line "line"
  = x: number s y: number s "l"
    { return { command: 'lineto', x, y } }
Curve "curve"
  = a: number s b: number s c: number s d: number s e: number s f: number s "c"
    { return { command: 'curveto', args: [a, b, c, d, e, f] } }
Home "home"
  = "h" { return { command: 'closepath' } }

s = " "

number "number"  = "-"? int frac? exp? { return parseFloat(text()) }
exp = [eE] ("+" / "-")? [0-9]+
frac = "." [0-9]+
int = "0" / ([1-9] [0-9]*)
```

carson-katri:
What do you think is left to be implemented for this to be working enough.

`Shape` is such a large part of SwiftUI that I personally think it would be best to implement it over time, instead of trying to create all of the functionality in one go. Also, some things are blocked by `Environment` and some sort of View metadata not being implemented. For instance, `foregroundColor` and `cornerRadius`.

Sources/TokamakCore/CGStubs.swift
```diff
@@ -15,4 +15,193 @@
...
+               [t2.tx, t2.ty, 1]]
+    var res = [[CGFloat]]()
+    for i in 0..<3 {
+      for j in 0..<3 {
+        res[i][j] = 0
+        for k in 0..<3 {
+          res[i][j] += t1m[i][k] * t2m[k][j]
+        }
+      }
+    }
```
MaxDesiatov, line 164: I think this might be a hot path with performance degrading because of array allocations. I don't know of optimizing it would be premature though, especially in absence of any benchmarks.

Sources/TokamakDemo/main.swift
```diff
@@ -37,6 +37,9 @@ let renderer = DOMRenderer(
     }
     VStack {
       Counter(count: 5, limit: 15)
+        .padding()
+        .background(Color(red: 0.9, green: 0.9, blue: 0.9, alpha: 1.0))
+        .border(Color.red, width: 3)
```
MaxDesiatov, line 42: Is there anything else in this PR that could be added to the demo to verify that things are working? (I know that absence of unit tests is not great, but I'm a bit blocked on resolving XCTest support on the toolchain side by https://github.com/swiftwasm/swift/issues/1330)

MaxDesiatov:
I am all for gradual implementations and small PRs, so if there isn't anything else to be added to the demo in this PR, I think this can be merged.

MaxDesiatov:
Also, is there anything here that could be added to `progress.md`?

Sources/TokamakDemo/main.swift
carson-katri: cornerRadius is implemented but only partially working...

carson-katri:
I’ll add a cornerRadius demo, and maybe try to switch the SVG demo to use shape?

MaxDesiatov:
If `cornerRadius` is not fully forking, is it worth including it in the demo?

carson-katri:
Maybe not. It really only applies the cornerRadius to the previous modifier (border, background, etc)

MaxDesiatov:
No worries, up to you if you'd like to modify the SVG demo, and it's probably only `progress.md` update left, unless @j-f1 has any other suggestions.

carson-katri:
I've added to `progress.md`. @MaxDesiatov  it's ready for review.

PR #122, opened 2020-06-30
Add Spacer support

carson-katri:
This is actually a bit more complex than it first seems.

1. If a Stack has a Spacer as a child, we need to fill the width or height of the parent so the Spacer's `flex-grow: 1;` works.
2. If a Stack has a child Stack along the cross axis (HStack with a nested VStack and vice vera) we need to fill the cross axis
3. If a Stack has a child Stack along the same axis (HStack with a nested HStack) we need to fill the axis.

This should be working correctly 😄

### DIFF
--- a/Sources/TokamakDOM/DOMRenderer.swift
+++ b/Sources/TokamakDOM/DOMRenderer.swift
@@ -57,6 +57,7 @@ public final class DOMRenderer: Renderer {
 
   public init<V: View>(_ view: V, _ ref: JSObjectRef) {
     rootRef = ref
+    rootRef.set("style", "display: flex; width: 100%; height: 100%; justify-content: center; align-items: center;")
     reconciler = StackReconciler(
       view: view,
       target: DOMNode(view, ref),
--- a/Sources/TokamakDOM/DOMRenderer.swift
+++ b/Sources/TokamakDOM/DOMRenderer.swift
@@ -57,7 +57,7 @@ public final class DOMRenderer: Renderer {
 
   public init<V: View>(_ view: V, _ ref: JSObjectRef) {
     rootRef = ref
-    rootRef.set("style", "display: flex; width: 100%; height: 100%; justify-content: center; align-items: center;")
+    rootRef.style = "display: flex; width: 100%; height: 100%; justify-content: center; align-items: center;"
     reconciler = StackReconciler(
       view: view,
       target: DOMNode(view, ref),
--- a/Sources/TokamakDOM/Views/Spacer.swift
+++ b/Sources/TokamakDOM/Views/Spacer.swift
@@ -1,0 +1,56 @@
+// Copyright 2020 Tokamak contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import TokamakCore
+
+public typealias Spacer = TokamakCore.Spacer
+
+enum SpacerContainerAxis {
+  case horizontal, vertical
+}
+
+protocol SpacerContainer: ParentView {
+  var hasSpacer: Bool { get }
+  var axis: SpacerContainerAxis { get }
+}
+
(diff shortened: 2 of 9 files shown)


Sources/TokamakDOM/Views/Spacer.swift
```diff
@@ -0,0 +1,69 @@
...
+
+protocol SpacerContainer: ParentView {
+  var hasSpacer: Bool { get }
+  var axis: SpacerContainerAxis { get }
+}
+
+extension SpacerContainer {
+  var hasSpacer: Bool {
+    children
+//      .filter { mapAnyView($0) { (v: Spacer) in v } != nil }
```
MaxDesiatov, line 31: **chore:** Shouldn't this be cleaned up?

Sources/TokamakDOM/Views/Spacer.swift
```diff
@@ -0,0 +1,69 @@
...
+        } != nil
+      }
+      .count > 0 ||
+      children.compactMap {
+        mapAnyView($0) { (v: SpacerContainer) in
+          v
+        }
+      }
+      .filter { $0.axis == axis }
+      .filter(\.hasSpacer)
```
MaxDesiatov, line 44: Slight optimization, but may be worth it.

```suggestion
      .filter { $0.axis == axis && $0.hasSpacer }
```

Sources/TokamakDOM/Views/Spacer.swift
```diff
@@ -0,0 +1,69 @@
...
+
+extension SpacerContainer {
+  var hasSpacer: Bool {
+    children
+//      .filter { mapAnyView($0) { (v: Spacer) in v } != nil }
+      .filter {
+        mapAnyView($0) { (v: Spacer) in
+          v
+        } != nil
+      }
```
MaxDesiatov, line 36: Can `compactMap` be used here consistently?

```suggestion
      .compactMap {
        mapAnyView($0) { (v: Spacer) in v }
      }
```

Sources/TokamakDOM/Views/Spacer.swift
```diff
@@ -0,0 +1,69 @@
...
+  // (e.g., an HStack with a child VStack which contains a spacer)
+  // If so, we need to fill the cross-axis so the child can show the correct layout.
+  var fillCrossAxis: Bool {
+    children
+      .compactMap {
+        mapAnyView($0, transform: { (v: SpacerContainer) in v })
+      }
+      .filter { $0.axis != axis }
+      .map(\.hasSpacer)
+      .filter { $0 }
```
MaxDesiatov, line 58:
```suggestion
      .filter { $0.axis != axis && $0.hasSpacer }
```

Sources/TokamakDOM/Views/Spacer.swift
```diff
@@ -0,0 +1,69 @@
...
+      .count > 0
+  }
+
+  // Does a child SpacerContainer along the opposite axis have a spacer?
+  // (e.g., an HStack with a child VStack which contains a spacer)
+  // If so, we need to fill the cross-axis so the child can show the correct layout.
+  var fillCrossAxis: Bool {
+    children
+      .compactMap {
+        mapAnyView($0, transform: { (v: SpacerContainer) in v })
```
MaxDesiatov, line 54: **nitpick:** This could be more consistent with the trailing closures uses

```suggestion
        mapAnyView($0) { (v: SpacerContainer) in v }
```

carson-katri:
I found a fix, however it requires adding styles to the root DOM node:

```css
display: flex; width: 100%; height: 100%; justify-content: center; align-items: center;
```

Those styles also make the initial layout of a simple "Hello World" match SwiftUI. We now will need a `ScrollView`, however.

Sources/TokamakDOM/DOMRenderer.swift
```diff
@@ -57,6 +57,7 @@ public final class DOMRenderer: Renderer {
 
   public init<V: View>(_ view: V, _ ref: JSObjectRef) {
     rootRef = ref
+    rootRef.set("style", "display: flex; width: 100%; height: 100%; justify-content: center; align-items: center;")
```
MaxDesiatov, line 60:
```suggestion
    rootRef.style = "display: flex; width: 100%; height: 100%; justify-content: center; align-items: center;")
```

MaxDesiatov:
Please also update `docs/progress.md`

carson-katri:
✅ or 🚧

MaxDesiatov:
This depends on whether we have any API and behavior differences left for it in this PR? 🙂

carson-katri:
There shouldn't be any API differences. Small behavior differences may be found through more heavy testing, but for now it all works as intended.

PR #125, opened 2020-06-30
Add TextFieldStyle

### LINKED ISSUE #121: Add style support
Many views support some kind of `.fooStyle()` method that takes one of a variety of `XFooStyle()` objects. I took a quick look and it seems like this kind of thing has not been implemented for any of the views. Ideally having one implementation of this would provide a base from which to implement similar methods for other views.

j-f1:
This didn’t end up being a modifier since that applies attributes to a wrapping `<div>` which doesn’t work for styles. Fixes #121.


j-f1:
Another thought: Is Semantic UI a good fit for this? It’s pretty much unmaintained and it has a bunch of CSS rules that get in the way of using regular semantic HTML elements to create styles.

Sources/TokamakCore/Modifiers/TextFieldStyle.swift
```diff
@@ -0,0 +1,65 @@
...
+  var type: String { get }
+  var style: String { get }
+  init()
+}
+
+public struct DefaultTextFieldStyle: TextFieldStyle {
+  public init() {}
+  public let type = "text"
+  public let style = """
+      -webkit-appearance: textfield;
```
carson-katri, line 29: I believe all the CSS stuff should be in TokamakDOM

MaxDesiatov:
I'm ready to swap Semantic UI for something more modern, just thought that Bootstrap is not a good option either, Material design is too Google-ish, but didn't find anything better. Any suggestions?

Turns out Danger can't get the token (or any secrets) for PRs from forked repositories, I'll disable it for forked repos for now.

Sources/TokamakCore/Modifiers/TextFieldStyle.swift
```diff
@@ -0,0 +1,65 @@
...
+//
+
+public protocol TextFieldStyle {
+  // FIXME: Should properties be internal?
+  var type: String { get }
+  var style: String { get }
+  init()
+}
+
+public struct DefaultTextFieldStyle: TextFieldStyle {
```
MaxDesiatov, line 25: I agree it would be best to keep those internal, how's about this?

```suggestion
public protocol TextFieldStyle {
}

protocol _TextFieldStyle: TextFieldStyle {
  var type: String { get }
  var style: String { get }
  init()
}

public struct DefaultTextFieldStyle: _TextFieldStyle {
```

MaxDesiatov:
I'm probably fine with Boostrap actually if there are no other options, my only excuse for avoiding it was that its default style is widely overused 🙈

j-f1:
Why not just go with the native browser styles? That way SwiftUI on the Mac, for example, would exactly match Tokamak on Safari.

MaxDesiatov:
Yeah, I hope that could work, another excuse is that Semantic UI was added in the demo, just as an example of adding custom CSS, but isn't enforced anywhere. @carson-katri do you have any opinion on this?

carson-katri:
I agree, almost all large CSS-libraries are fairly opinionated. The user could include Bootstrap, Material, etc. themselves, and then if they already have styling and are embedding Tokamak, it will match.

MaxDesiatov:
Do you think it's still worth it keeping that Semantic UI snippet in the demo then?

carson-katri:
It has minimal effect as we're not using their class names anywhere. I think it'd be more beneficial once we have a way for Tokamak to insert custom class names on Views.

Sources/TokamakCore/Modifiers/TextFieldStyle.swift
```diff
@@ -0,0 +1,65 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//  Created by Jed Fox on 06/30/2020.
+//
+
+public protocol TextFieldStyle {
+  // FIXME: Should properties be internal?
+  var type: String { get }
```
carson-katri, line 20: These two properties seems very TokamakDOM specific.

Could you make this protocol empty, and extend it in TokamakDOM?

Here's what SwiftUIs `TextFieldStyle` looks like if you'd want to match that
```swift
public protocol TextFieldStyle {
  associatedtype _Body : SwiftUI.View
  func _body(configuration: SwiftUI.TextField<Self._Label>) -> Self._Body
  typealias _Label = SwiftUI._TextFieldStyleLabel
}

public struct _TextFieldStyleLabel : SwiftUI.View {
  public typealias Body = Swift.Never
}
```

Sources/TokamakCore/Modifiers/TextFieldStyle.swift
```diff
@@ -0,0 +1,65 @@
...
+  public init() {}
+  public let type = "text"
+  public let style = """
+      -webkit-appearance: textfield;
+      appearance: textfield;
+  """
+}
+
+extension TextField where Label == Text {
+  public func textFieldStyle(_ style: TextFieldStyle) -> Self {
```
carson-katri, line 62: If you match the SwiftUI API you'll have to make this generic.

carson-katri: In SwiftUI, `textFieldStyle` is available on any view, and should effect all child views if I'm not mistaken:

```swift
extension View {
  public func textFieldStyle<S>(_ style: S) -> some SwiftUI.View where S : SwiftUI.TextFieldStyle
}
```

This may not be possible until we get `Environment` working, however.

MaxDesiatov:
@carson-katri I concur 👍

Sources/TokamakCore/Modifiers/TextFieldStyle.swift
j-f1: Will move. FWIW, the `appearance` override is only necessary because Semantic UI sets it to `none`.

Sources/TokamakCore/Modifiers/TextFieldStyle.swift
j-f1: I think this works well enough for now and fits most use cases, and it can be expanded once `Environment`/context is available

Sources/TokamakCore/Modifiers/TextFieldStyle.swift
j-f1: I’m not exactly sure how that works. How are you accessing these headers? (so I can look at them in the future)

carson-katri: They're located inside of Xcode: `/Applications/Xcode-beta/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/SwiftUI.framework/Modules/SwiftUI.swiftmodule/arm64.swiftinterface`

Super useful, especially when something is marked `@inlinable`

carson-katri: As for how it works, all of these `XStyle` types seems to contain a `body` function with a configuration (which contains the label usually) passed in. This lets the user apply modifiers to the label to define the style of the component.

j-f1: aha I think I see what happens. The TextFieldStyle is wrapped around every TextField beneath it somehow by calling `someTextFieldStyle._body(textField)`, the output of which is swapped in for the actual `TextField`. Weirdly, `DefaultTextFieldStyle` is implemented differently, returning a custom view that wraps the actual body.

```swift
public struct RoundedBorderTextFieldStyle : SwiftUI.TextFieldStyle {
  public init()
  public func _body(configuration: SwiftUI.TextField<SwiftUI.RoundedBorderTextFieldStyle._Label>) -> some SwiftUI.View
  
  public typealias _Body = /* opaque return type */
}

public struct DefaultTextFieldStyle : SwiftUI.TextFieldStyle {
  public init()
  public func _body(configuration: SwiftUI.TextField<SwiftUI.DefaultTextFieldStyle._Label>) -> SwiftUI.DefaultTextFieldStyle._Body
  public struct _Body : SwiftUI.View {
    public var body: some SwiftUI.View { get }
    public typealias Body = /* opaque return type */
  }
}
```

Sources/TokamakCore/Modifiers/TextFieldStyle.swift
```diff
@@ -0,0 +1,42 @@
...
+  public init() {}
+}
+
+public struct SquareBorderTextFieldStyle: TextFieldStyle {
+  public init() {}
+}
+
+extension TextField where Label == Text {
+  public func textFieldStyle(_ style: TextFieldStyle) -> Self {
+    _TextFieldProxy(self).with(style: style).subject
```
MaxDesiatov, line 40: **suggestion:** No need to use a proxy within `TokamakCore` as you have access to all these properties and types directly. The only purpose of proxies is to give access to internal properties to modules other than `TokamakCore`, like `TokamakDOM`.

```suggestion
    TextField(from: self, textFieldStyle: style)
```

Sources/TokamakCore/Views/TextField.swift
```diff
@@ -33,9 +34,21 @@ extension TextField where Label == Text {
...
     label = Text(title)
-    textBinding = text.projectedValue
+    textBinding = text
     self.onEditingChanged = onEditingChanged
     self.onCommit = onCommit
+    textFieldStyle = DefaultTextFieldStyle()
+  }
+
+  init(
+    _from textField: TextField,
```
MaxDesiatov, line 44: **nitpick:** Not sure if underscore here is supposed to indicate that something is internal?

```suggestion
    from textField: TextField,
```

Sources/TokamakCore/Views/TextField.swift
```diff
@@ -52,8 +65,13 @@ public struct _TextFieldProxy {
 
   public init(_ subject: TextField<Text>) { self.subject = subject }
 
+  public func with(style: TextFieldStyle) -> Self {
+    _TextFieldProxy(TextField(_from: subject, textFieldStyle: style))
+  }
+
```
MaxDesiatov, line 71: If you apply suggestions above, there's no need for this function.

```suggestion
  public func with(style: TextFieldStyle) -> Self {
    _TextFieldProxy(TextField(_from: subject, textFieldStyle: style))
  }

```

Sources/TokamakDemo/main.swift
```diff
@@ -17,11 +17,11 @@ import TokamakDOM
...
-_ = document.head.object!.insertAdjacentHTML!("beforeend", #"""
-<link
-  rel="stylesheet"
-  href="https://cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.4.1/semantic.min.css">
-"""#)
+// _ = document.head.object!.insertAdjacentHTML!("beforeend", #"""
+// <link
+//  rel="stylesheet"
+//  href="https://cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.4.1/semantic.min.css">
+// """#)
```
MaxDesiatov, line 24: Let's remove it altogether if it's now unused.

```suggestion
// _ = document.head.object!.insertAdjacentHTML!("beforeend", #"""
// <link
//  rel="stylesheet"
//  href="https://cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.4.1/semantic.min.css">
// """#)
```

Sources/TokamakCore/Views/TextField.swift
j-f1: Yep! That's the idea since you're not supposed to be able to pass a TextFieldStyle directly into a TextField.

MaxDesiatov: The function is not `public` though, so nothing outside of `TokamakCore` can use it, which falls in the area of ["rule 1"](https://github.com/swiftwasm/Tokamak/pull/137/files#diff-04c6e90faac2675aa89e2176d2eec7d8R126) I attempted to describe in #137, but I'll let you review those rules first 🙂

j-f1:
I noticed [this line](https://github.com/SwiftWebUI/SwiftWebUI/blob/16b84d46dc4ea86da335440a0979a3cff0daa9f3/Sources/SwiftWebUI/Views/Forms/TextField.swift#L9) in SwiftWebUI:

```swift
import class Foundation.Formatter
```

Is this a way to include Foundation classes without making the bundle too large or does it not tree-shake at all yet?

MaxDesiatov:
To my knowledge imports like this only limit the symbol resolution scope, but don't import just a subset of a module as there are many things that you don't explicitly import but come implicitly like extensions, protocol conformances etc.

There's no dead code elimation for unused protocol conformances, and while Foundation doesn't seem to be particularly protocol-heavy, in my testing it adds about 10 MB to the final binary. Maybe there's still some DCE that can be done even with something like Foundation, in my understanding this is what [the GSoC LTO project](https://forums.swift.org/t/gsoc-lto-support-progress-report/37149) is about.

Also, Foundation heavily relies on CoreFoundation, which is written in C, maybe that's the reason too, but this is only my speculation.

Maybe it's still worth trying to link it and re-evaluating whether it's worth it, for me the blocker is in `carton` right now: https://github.com/swiftwasm/carton/issues/4

carson-katri:
Here's a great article on importing declarations: https://nshipster.com/import/

In SwiftWebUI it's not used for disambiguation as much as clarity.

Sources/TokamakCore/Modifiers/TextFieldStyle.swift
```diff
@@ -0,0 +1,42 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//  Created by Jed Fox on 06/30/2020.
+//
+
+public protocol TextFieldStyle {
```
carson-katri, line 18: Since this protocol is public in SwiftUI, I think its important to match:

```swift
public protocol TextFieldStyle {
  associatedtype _Body : SwiftUI.View
  func _body(configuration: SwiftUI.TextField<Self._Label>) -> Self._Body
  typealias _Label = SwiftUI._TextFieldStyleLabel
}
```

But we could handle this in a later PR if you'd prefer...

Sources/TokamakCore/CGStubs.swift
```diff
@@ -15,7 +15,11 @@
 //  Created by Max Desiatov on 08/04/2020.
 //
 
+#if canImport(Glibc)
```
MaxDesiatov, line 18: Is there anything in this file that requires Foundation?

j-f1:
ok now we’re actually good to go. I went with `\.textFieldStyle🔒` so users would still be able to specify their own `textFieldStyle` environment if they wanted but I could also drop the 🔒. Maybe there could be a separate parallel private environment only available to Tokamak for storing internal state like this?

Sources/TokamakCore/CGStubs.swift
j-f1: No, but Xcode doesn’t like the Glibc import and Foundation includes the necessary trig functions. Is there something else that should be imported in native platforms?

j-f1: It doesn’t seem to affect the warm bundle size at all.

carson-katri: `import Darwin` for the trig functions.

MaxDesiatov: I'd suggest importing based on the platform then:

```suggestion
#if os(WASI)
```

Sources/TokamakCore/CGStubs.swift
```diff
@@ -15,7 +15,11 @@
...
 import Glibc
+#else
+import Foundation
+#endif
```
MaxDesiatov, line 22: Yeah, `import Darwin` could work better:
```suggestion
#elseif canImport(Darwin)
import Darwin
#endif
```

Sources/TokamakCore/CGStubs.swift
```diff
@@ -15,7 +15,11 @@
 //  Created by Max Desiatov on 08/04/2020.
 //
 
+#if os(WASI)
 import Glibc
+#else
+import Foundation
+#endif
```
MaxDesiatov, line 22: Sorry about changing my mind, here's what would be most cross-platform in my opinion:

```suggestion
#if canImport(Glibc)
import Glibc
#elseif canImport(Darwin)
import Darwin
#endif
```

j-f1: No worries!

j-f1:
I guess I have another question: should the `TextFieldStyle.swift` file go in `Modifiers` (since it has one) or `Environment` (since it has one too) or should I make a new folder for `Styles`?

MaxDesiatov:
Are there going to be multiple other styles? If so, separate `Styles` directory makes sense to me.

Sources/TokamakCore/Modifiers/TextFieldStyle.swift
```diff
@@ -0,0 +1,55 @@
...
+public struct SquareBorderTextFieldStyle: TextFieldStyle {
+  public init() {}
+}
+
+enum TextFieldStyleKey: EnvironmentKey {
+  static var defaultValue: TextFieldStyle = DefaultTextFieldStyle()
+}
+
+extension EnvironmentValues {
+  var textFieldStyle🔒: TextFieldStyle {
```
MaxDesiatov, line 41: Sorry, I'm not sure I'm following, why not just name it `textFieldStyle`? It's an internal symbol anyway, users won't have access to it, as far as I understand.

j-f1: I don’t think people would be able to create a custom `textFieldStyle` environment variable if this private one had that name. Not sure that’s an issue though, but that was my reasoning.

MaxDesiatov: They totally would be able to, when symbol names are compiled down to machine code (or WebAssembly for that matter) [they're mangled](https://mikeash.com/pyblog/friday-qa-2014-08-15-swift-name-mangling.html) with module names prefixing them, so even public symbols with the same name don't collide. Hope that makes sense 🙂

PR #128, opened 2020-06-30
Add ScrollView

carson-katri:
`ScrollView` is now possible (was definitely necessary with the new root styles). It matches the behavior of SwiftUI as closely as possible. For instance, the ScrollView won't fill the parent's width unless you put a

```swift
HStack {
  Spacer()
}
```

inside of it (`VStack` for horizontal scroll).

Currently scrollbars can only be shown/hidden on Firefox.

### DIFF
--- a/Sources/TokamakDOM/Views/ScrollView.swift
+++ b/Sources/TokamakDOM/Views/ScrollView.swift
@@ -31,12 +31,9 @@ extension ScrollView: ViewDeferredToRenderer, SpacerContainer {
   public var deferredBody: AnyView {
     let scrollX = axes.contains(.horizontal)
     let scrollY = axes.contains(.vertical)
-    // FIXME: I can only hide the scrollbars on Firefox
-    // Safari/Chrome require a `::-webkit-scrollbar` selector, which can't be done with inline styles
     return AnyView(HTML("div", ["style": """
       \(scrollX ? "overflow-x: auto; width: 100%;" : "overflow-x: hidden;")
       \(scrollY ? "overflow-y: auto; height: 100%;" : "overflow-y: hidden;")
-      \(!showsIndicators ? "scrollbar-color: transparent;" : "")
       \(fillCrossAxis && scrollX ? "height: 100%;" : "")
       \(fillCrossAxis && scrollY ? "width: 100%;" : "")
       """,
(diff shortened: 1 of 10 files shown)


Sources/TokamakDOM/Views/ScrollView.swift
```diff
@@ -0,0 +1,46 @@
...
+    } else {
+      return .vertical
+    }
+  }
+
+  public var deferredBody: AnyView {
+    let scrollX = axes.contains(.horizontal)
+    let scrollY = axes.contains(.vertical)
+    // FIXME: I can only hide the scrollbars on Firefox
+    // Safari/Chrome require a `::-webkit-scrollbar` selector, which can't be done with inline styles
```
j-f1, line 35: Could you make a global CSS rule like

```css
._tokamak-indicatorless-scrollview::-webkit-scrollbar {
  width: 0;
  height: 0;
}
._tokamak-indicatorless-scrollview::scrollbar {
  width: 0;
  height: 0;
}
```

carson-katri: Yeah, that would work. We currently don't inject any styles to the DOM, but we could add one. @MaxDesiatov Thoughts?

MaxDesiatov: Yeah, `DOMRenderer` would be a good entrypoint for that 👍

Sources/TokamakDOM/Views/ScrollView.swift
```diff
@@ -0,0 +1,44 @@
...
+  public var deferredBody: AnyView {
+    let scrollX = axes.contains(.horizontal)
+    let scrollY = axes.contains(.vertical)
+    return AnyView(HTML("div", ["style": """
+      \(scrollX ? "overflow-x: auto; width: 100%;" : "overflow-x: hidden;")
+      \(scrollY ? "overflow-y: auto; height: 100%;" : "overflow-y: hidden;")
+      \(fillCrossAxis && scrollX ? "height: 100%;" : "")
+      \(fillCrossAxis && scrollY ? "width: 100%;" : "")
+      """,
+                                "class": !showsIndicators ? "_tokamak-scrollview-hideindicators" : ""]) {
```
MaxDesiatov, line 40: Indentation here looks weird, you'll have to put the `"style"` key on a new line for it to auto-format properly.

carson-katri:
👍  will do for my next branch

PR #136, opened 2020-06-30
Add `ColorScheme` environment

MaxDesiatov:
Was temporarily blocked by #135.


ie-ahm-robox:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  **Sources/TokamakDOM/DOMRenderer.swift#L62** - Line should be 100 characters or less: currently 127 characters (`line_length`)
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against 34545ed97c6f41bb21a5f15b52f052d2ea692700
</p>

Sources/TokamakCore/Tokens/ColorScheme.swift
```diff
@@ -0,0 +1,39 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+public enum ColorScheme: CaseIterable {
+  case dark
+  case light
+}
+
+struct ColorSchemeKey: EnvironmentKey {
+  static let defaultValue: ColorScheme = .light
```
j-f1, line 21: Would it be possible for this to match the value of the `(prefers-color-scheme: dark)` media query? You could create the query with `window.matchMedia('(prefers-color-scheme: dark)')` and add an event listener to update this value on change.

MaxDesiatov: Yes, that's what I'm working right now. I think it would make sense to add the listener in an `onAppear` closure on the new `DOMEnvironment` view introduced in this PR, but this obviously needs a working `onAppear` implementation first, which is what I'm working on right now 🙂

Sources/TokamakDOM/Environment/DOMEnvironment.swift
```diff
@@ -0,0 +1,44 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import JavaScriptKit
+import TokamakCore
+
+/// This is effectively a singleton, but a mounted `DOMEnvironment` is assumed to be a singleton
+/// too. It can't be declared as a property of `DOMEnvironment` as you can't modify it from within
+/// `body` callbacks.
+private var colorSchemeListener: JSClosure?
```
j-f1, line 21: Could you make it a `@State` then initialize it in the struct like so?


```swift
@State var colorSchemeListener = JSClosure {
  scheme = $0[0].object!.matches.boolean == true ? .dark : .light
  return .undefined
}
````

Comment (MaxDesiatov) [reply Sources/TokamakDOM/Environment/DOMEnvironment.swift:21]: I tried that, but every assignment on a `@State` property triggers a needless re-render, while this assignment to a private global doesn't.

**Comment by MaxDesiatov** [comment]
Yes, the listener works, but environment changes aren't propagated properly yet, I'm looking into that now.

File: Sources/TokamakDOM/Environment/DOMEnvironment.swift (hunk 2 above)
Comment (j-f1) [reply Sources/TokamakDOM/Environment/DOMEnvironment.swift:21]: Yeah, but if you only assign it in the definition of the struct it won’t change on every render, right?

Comment (j-f1) [reply Sources/TokamakDOM/Environment/DOMEnvironment.swift:21]: oh nope nvm

Comment (MaxDesiatov) [reply Sources/TokamakDOM/Environment/DOMEnvironment.swift:21]: Ah, I see what you mean. That won't work because you're referencing `self` (implicitly via the reference to `scheme`) in the variable definition, `self` is not fully initialized yet at that point, which causes this error:

```
error: instance member 'scheme' cannot be used on type 'DOMEnvironment<V>'; 
did you mean to use a value of this type instead?
    scheme = $0[0].object!.matches.boolean == true ? .dark : .light
    ^~~~~~
```

**Comment by carson-katri** [comment]
Is this unblocked since #170 ?

**Comment by MaxDesiatov** [comment]
I'm still seeing some issues with environment propagation (at least in this PR), working on that right now...

File: Sources/TokamakDemo/EnvironmentDemo.swift [hunk 3]
```diff
@@ -32,15 +32,31 @@ struct EnvironmentObjectDemo: View {
   }
 }
 
+extension ColorScheme: CustomStringConvertible {
+  public var description: String {
+    switch self {
+    case .dark: return "dark"
+    case .light: return "light"
+        @unknown default:
```
ie-ahm-robox, line 40: - :no_entry_sign: Case statements should vertically align with their enclosing switch statement. (`switch_case_alignment`)

j-f1:
If you’re going to add support for dark mode it should also add the CSS property `color-scheme: light dark` to the html or body element. Not sure if there’s a way to forcibly enable or disable dark mode support in swiftui, but it should set it to `light only` or `dark only` if so.

carson-katri:
I believe you would do that with `environment(\.colorScheme, .light)`

MaxDesiatov:
Good point. It still doesn't work because there's apparently a bug somewhere in `JSClosure` implementation in JavaScriptKit which makes the color scheme observer overwrite the scene phase observer, both can't exist at the same time. I'm trying to isolate it now and this is what led me to the attempt at reviving our test suite in #229. Now I see the test suite can't run on WASI, I'm back to looking into the test suite of JavaScriptKit...

MaxDesiatov:
I've fixed the issue with observers overwriting each other's closures. Turns out we have to be very careful with the lifetime of `JSClosure`. If an instance of it is not retained by the time next `JSClosure` instance is created, their `ObjectIdentifier` values can become the same, which will lead to hard to diagnose issues. PR to `JavaScriptKit` is coming soon.

I'm still seeing what seems like yet another environment propagation issue, but it's only reproducible with the environment not going as far as `EnvironmentDemo`, views closer to the root get `colorScheme` appropriately when the system scheme changes.

MaxDesiatov:
The last known environment update issue is fixed and it's ready for review. You can see the dark scheme environment text representation updated in `EnvironmentDemo`. As it has been open for a while now, I suggest adding default dark mode styles in a separate PR, I've created #237 as a reminder for that.

j-f1:
I gave it a test and it seems like clicking the dark mode toggle in Safari Developer Tools does not work unless you click it before the first render — subsequent changes are not reflected in the view.

MaxDesiatov:
I was testing with the globally set dark mode in System Preferences, do you see the same behavior with that one?

j-f1:
Yes, I see the same behavior there

MaxDesiatov:
Hm, I had this same issue caused by the environment bug, which was fixed in c397eb6c13ec3945fd1618c07de87f4a633589dd. Maybe you're not on the latest commit of this branch?

carson-katri:
It's working for me in Chrome. Maybe a bug in Safari?

Sources/TokamakCore/App/_AnyApp.swift
```diff
@@ -48,7 +48,11 @@ public struct _AnyApp: App {
     fatalError("`title` cannot be set for `AnyApp`. Access underlying `app` value.")
   }
 
-  public var _phasePublisher: CurrentValueSubject<ScenePhase, Never> {
+  public var _phasePublisher: AnyPublisher<ScenePhase, Never> {
+    fatalError("`_AnyApp` cannot monitor scenePhase. Access underlying `app` value.")
+  }
+
+  public var _colorSchemePublisher: AnyPublisher<ColorScheme, Never> {
     fatalError("`_AnyApp` cannot monitor scenePhase. Access underlying `app` value.")
```
carson-katri, line 56: The message here should say `colorScheme`

Sources/TokamakDemo/EnvironmentDemo.swift
```diff
@@ -32,15 +32,30 @@ struct EnvironmentObjectDemo: View {
...
     VStack {
-      Text(font == nil ? "`font` environment not set." : "\(String(describing: font!))")
-      Text(testEnv.envTest)
-      EnvironmentObjectDemo()
+      Text("`colorScheme` is \(colorScheme.description)")
+      if let font = font {
+        Text("`font` environment is \(String(describing: font))")
+      }
+      // Text(testEnv.envTest)
+      // EnvironmentObjectDemo()
```
carson-katri, line 58: Any reason why the `EnvironmentObjectDemo` was disabled?

Sources/TokamakCore/App/App.swift
```diff
@@ -32,7 +32,10 @@ public protocol App: _TitledApp {
   static func _launch(_ app: Self, _ rootEnvironment: EnvironmentValues)
 
   /// Implemented by the renderer to update the `App` on `ScenePhase` changes
-  var _phasePublisher: CurrentValueSubject<ScenePhase, Never> { get }
+  var _phasePublisher: AnyPublisher<ScenePhase, Never> { get }
```
carson-katri, line 35: These need to be implement in TokamakStaticHTML to fix the build.

Sources/TokamakDemo/EnvironmentDemo.swift
MaxDesiatov: Was just experimenting with it, but forgot to uncomment, sorry.

Sources/TokamakStaticHTML/Core.swift
```diff
@@ -55,6 +76,8 @@ public typealias CGSize = TokamakCore.CGSize
 
 // MARK: Views
 
+public typealias Button = TokamakCore.Button
```
carson-katri, line 79: I intentionally left out things like `Button` and `TextField` as they're pretty much useless without `State`. Do you think we should include them?

carson-katri: What we can add which will make TokamakStaticHTML more useful is [`Link`](https://developer.apple.com/documentation/swiftui/link)

MaxDesiatov: I wonder if we could still render stateful `View`s to static HTML, any listeners on those would be just ignored. But that would obviously come as a separate PR. LMK if you think that these should be removed in this PR.

carson-katri: I think they should be removed, plus they only have HTML bodies in TokamakDOM.

PR #142, opened 2020-07-01
Add Xcode project and native targets

j-f1:
This makes the demo run on iOS and macOS. I slimmed down the Xcode project as much as possible to keep it light.


Tokamak Native/UIAppDelegate.swift
```diff
@@ -0,0 +1,33 @@
...
+import UIKit
+
+// so we only need one Info.plist
+public class NSApplication: UIApplication {}
+
+@UIApplicationMain
+class AppDelegate: UIResponder, UIApplicationDelegate {
+  let window = UIWindow()
+  func application(_ application: UIApplication,
+                   didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil) -> Bool {
```
ie-ahm-robox, line 28: - :no_entry_sign: Line should be 100 characters or less: currently 119 characters (`line_length`)

carson-katri:
You could probably slim it down even more if you use the SwiftUI lifecycle

j-f1:
I could but I don't think apps written with it will run on non-beta OSes atm

Sources/TokamakCore/Tokens/Color.swift
```diff
@@ -24,26 +24,26 @@ public struct Color: Equatable {
   public let red: Double
   public let green: Double
   public let blue: Double
-  public let alpha: Double
+  public let opacity: Double
```
MaxDesiatov, line 27: Woah, this uncovers a lot of incompatibilities with SwiftUI, looks like this type is again vestigial from the React implementation... No need to modify it in this PR though, just a note to self 🙂

Tokamak Native/NSAppDelegate.swift
```diff
@@ -0,0 +1,36 @@
+// Copyright 2019-2020 Tokamak contributors
```
MaxDesiatov, line 1: How's about renaming the directory to `TokamakDemo Native` as it only compiles the demo code?

Sources/TokamakCore/Tokens/Color.swift
j-f1: I had to modify it in this PR because there’s a call to `Color` in the demo code which wouldn’t compile for SwiftUI without this change.

MaxDesiatov: I understand, thanks for cleaning that up.

MaxDesiatov:
The build is failing with this error:

```
/github/workspace/Sources/TokamakDOM/Tokens/Tokens.swift:22:57: error: cannot find 'alpha' in scope
    "rgb(\(red * 255), \(green * 255), \(blue * 255), \(alpha * 255))"
```

PR #145, opened 2020-07-02
Add AppearanceActionModifier, onAppear/onDisappear

MaxDesiatov:
This is required to unblock #136 as I think it would make sense for the `DOMEnvironment` view there to add/remove its color scheme listener in `onAppear`/`onDisappear` closures.

I've also restored full SwiftUI compatibility in the signature of `func modifier<Modifier>(_ modifier: Modifier)`. Consequently `ViewDeferredToRenderer` had to be implemented on `ModifiedContent` then instead of `_ViewModifier_Content` to make it work.

### DIFF
--- a/Sources/TokamakCore/Modifiers/AppearanceActionModifier.swift
+++ b/Sources/TokamakCore/Modifiers/AppearanceActionModifier.swift
@@ -1,0 +1,45 @@
+// Copyright 2020 Tokamak contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+protocol AppearanceActionProtocol {
+  var appear: (() -> ())? { get }
+  var disappear: (() -> ())? { get }
+}
+
+struct _AppearanceActionModifier: ViewModifier {
+  var appear: (() -> ())?
+  var disappear: (() -> ())?
+  init(appear: (() -> ())? = nil, disappear: (() -> ())? = nil) {
+    self.appear = appear
+    self.disappear = disappear
+  }
+
+  typealias Body = Never
+}
+
+extension ModifiedContent: AppearanceActionProtocol
+  where Content: View, Modifier == _AppearanceActionModifier {
+  var appear: (() -> ())? { modifier.appear }
+  var disappear: (() -> ())? { modifier.disappear }
+}
+
+extension View {
+  public func onAppear(perform action: (() -> Swift.Void)? = nil) -> some View {
+    modifier(_AppearanceActionModifier(appear: action, disappear: nil))
+  }
+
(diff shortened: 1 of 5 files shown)


Sources/TokamakCore/Modifiers/AppearanceActionModifier.swift
```diff
@@ -0,0 +1,46 @@
...
+
+extension ModifiedContent: AppearanceActionProtocol
+  where Content: View, Modifier == _AppearanceActionModifier {
+  var appear: (() -> ())? { modifier.appear }
+  var disappear: (() -> ())? { modifier.disappear }
+}
+
+extension View {
+  public func onAppear(perform action: (() -> Swift.Void)? = nil) -> some View {
+    modifier(_AppearanceActionModifier(appear: action, disappear: nil))
```
j-f1, line 40:
```suggestion
     modifier(_AppearanceActionModifier(appear: action))
```

And same for the one below. Alternatively, you don’t need the custom initializer and can use the native struct initializer.

Sources/TokamakCore/MountedViews/MountedCompositeView.swift
```diff
@@ -44,6 +44,10 @@ final class MountedCompositeView<R: Renderer>: MountedView<R>, Hashable {
   override func mount(with reconciler: StackReconciler<R>) {
     let childBody = reconciler.render(compositeView: self)
 
+    if let appearanceAction = view.view as? AppearanceActionProtocol {
```
j-f1, line 47: Will this work if there's another modifier wrapped around the appearance action modifier?

Sources/TokamakDemo/Counter.swift
```diff
@@ -29,6 +29,8 @@ public struct Counter: View {
           Button("Increment") { count += 1 }
           Text("\(count)")
         }
+        .onAppear { print("Counter.VStack onAppear") }
+        .onDisappear { print("Counter.VStack onDisappear") }
```
j-f1, line 33: Where do `print` statements get outputted?

Sources/TokamakCore/MountedViews/MountedCompositeView.swift
MaxDesiatov: I think it does work, at least in my testing where both `onAppear` and `onDisappear` are chained. The wrapping modifier will get this function called separately. In my understanding wrapped modifiers are peeled like an onion by this function recursively.

Sources/TokamakDemo/Counter.swift
MaxDesiatov: In Web Inspector console, all `print` output is redirected with `console.log`.

Sources/TokamakCore/Modifiers/AppearanceActionModifier.swift
```diff
@@ -0,0 +1,46 @@
...
+}
+
+extension ModifiedContent: AppearanceActionProtocol
+  where Content: View, Modifier == _AppearanceActionModifier {
+  var appear: (() -> ())? { modifier.appear }
+  var disappear: (() -> ())? { modifier.disappear }
+}
+
+extension View {
+  public func onAppear(perform action: (() -> Swift.Void)? = nil) -> some View {
```
carson-katri, line 39: I'm guessing this came from the `swiftinterface` file, but `Swift.Void` can just be `Void`, I personally find that to be cleaner.

MaxDesiatov: Oh, wasn't intentional, gonna fix it now, great shout

PR #160, opened 2020-07-05
Use `let` instead of `var` in `TextFieldStyleKey: EnvironmentKey`

MaxDesiatov:
`static var` makes it effectively a global variable, while it should be a constant.


MaxDesiatov:
No worries, this occasionally gets me too 🙂

PR #162, opened 2020-07-05
ViewBuilder buildIf fix

carson-katri:
This allows `Optional` views to be rendered, fixing the `buildIf` function on `ViewBuilder`.

Ideally, type erasure would be unnecessary, but this causes a crash (see https://github.com/swiftwasm/swift/issues/1379).

`List` implementation will be much simpler once this is resolved. Maybe we can merge this for now, then take out the `AnyView` stuff when the swift issue is resolved?

### DIFF
--- a/Sources/TokamakCore/Views/ViewBuilder.swift
+++ b/Sources/TokamakCore/Views/ViewBuilder.swift
@@ -45,6 +45,8 @@ public struct _ConditionalContent<TrueContent, FalseContent>: View
   }
 }
 
+// FIXME: Remove type erasure when https://github.com/swiftwasm/swift/issues/1379
+// is resolved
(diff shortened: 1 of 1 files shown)


Sources/TokamakCore/Views/ViewBuilder.swift
```diff
@@ -25,13 +25,33 @@ public struct EmptyView: View {
...
+    case let .trueContent(view):
+      view
+    case let .falseContent(view):
+      view
+    }
+  }
+}
+
+extension Optional: View where Wrapped: View {
+  public var body: AnyView {
```
MaxDesiatov, line 49: Please add a FIXME that referers to the relevant issue, much better chance we don't forget about it then.

PR #171, opened 2020-07-09
Add @ObservedObject implementation

MaxDesiatov:
This pulls a fork of OpenCombine that can be compiled with the same SwiftWasm snapshot we use in `main`.

The only caveat is that this doesn't work for `ObservableObject`s that are subclasses of other classes with `@Published` properties. This issue needs to be fixed in [the Runtime library fork](https://github.com/MaxDesiatov/Runtime). Since this is a rare case, and fixing it wouldn't change this `@ObservedObject` implementation, I think it's ready for review as is.


Sources/TokamakDOM/Shapes/Path.swift
```diff
@@ -50,7 +50,7 @@ extension Path: ViewDeferredToRenderer {
         "rx": "\(roundedRect.cornerSize.width)",
         "ry": "\(roundedRect.style == .continuous ? roundedRect.cornerSize.width : roundedRect.cornerSize.height)",
```
ie-ahm-robox, line 51: - :no_entry_sign: Line should be 100 characters or less: currently 115 characters (`line_length`)

ie-ahm-robox:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  **Sources/TokamakDOM/Shapes/Path.swift#L57** - Line should be 100 characters or less: currently 113 characters (`line_length`)
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against 3a41603a2b932560e6f45ec7a9e35a95111d1ffa
</p>

Sources/TokamakDOM/Shapes/Path.swift
ie-ahm-robox, line 51: - :no_entry_sign: Line should be 100 characters or less: currently 115 characters (`line_length`)

ie-ahm-robox:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  **Sources/TokamakDOM/Shapes/Path.swift#L57** - Line should be 100 characters or less: currently 113 characters (`line_length`)
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against b9da18a101a3de90aeca450d655004f045e465ff
</p>

carson-katri:
What is left in OpenCombine for this to be working

MaxDesiatov:
In my current investigation, I'm not entirely convinced it's a problem in OpenCombine (at least not in its high-level code), this is either a problem in SwiftWasm runtime, or the way that COpenCombineHelpers processes runtime metadata when looking for the `@Published` property. I also tried to swap that C++ implementation to [the Runtime library](https://github.com/MaxDesiatov/Runtime), but got similar errors. I've also tried running the Runtime library test suite, which currently fails with classes, I don't remember seeing those test failures a couple of months ago. I've got `carton test` implemented in the `main` branch of `carton` and currently working on fixing those test failures. If the OpenCombine issues persist after those tests are fixed, I think we should isolate that failure and add as a test case to avoid breakage in the future.

MaxDesiatov:
TWIMC, I'm currently looking at [the type metadata docs](https://github.com/apple/swift/blob/master/docs/ABI/TypeMetadata.rst), looks like both Runtime library and COpenCombineHelpers have problems particulalrly with classes metadata, even when classes aren't generic. These issues are reproducible only with SwiftWasm, I'm not sure if this is because of some 64-bit platform assumptions in their code, or due to some issue we have in SwiftWasm. Since the rest of the things in SwiftWasm work pretty well in my knowledge, I think it's more about the runtime introspection code in those libraries, but it's not that easy to debug...

MaxDesiatov:
> On platforms which support Objective-C interoperability, two words are
  reserved for use by the Objective-C runtime at **offset 2** and **offset
  3**; on other platforms, nothing is reserved.
>- On platforms which support Objective-C interoperability, the **rodata 
  pointer** is stored at **offset 4**; on other platforms, it is not present. 

That could be it, but I thought that both of those libraries work well on Linux, which doesn't have Objective-C interop 🤔

MaxDesiatov:
Given that https://github.com/apple/swift/pull/32374 was merged only recently, they might have changed metadata layout on non-Apple platforms recently as they don't need to maintain ABI stability on those platforms.

MaxDesiatov:
This no longer needs any custom or new toolchains, the original one works fine 🙂

MaxDesiatov:
@carson-katri @j-f1 after this one's merged I think I'm going to tag Tokamak 0.2.0 as is, so then `carton init` template introduced in https://github.com/swiftwasm/carton/pull/54 could specify 0.2.0 version instead of the `main` branch. I'd prefer users specifying a specific version in their package manifests to avoid unexpected breakage. Additionally, if any 3rd-party library depending on Tokamak appears, if both user's application and that library dependn on specific Tokamak version, it's much easier to resolve the dependency tree, but harder if either of those depend on a branch.

MaxDesiatov:
Conflicts resolved now.

PR #174, opened 2020-07-11
Add Text concatenation via + operator

carson-katri:
```swift
Text("This text is ") + Text("concatenated").bold()
```
results in:

> This text is **concatenated**

```swift
(Text("This text is ") + Text("concatenated").bold())
  .italic()
```
results in:

> *This text is **concatenated***

### DIFF
--- a/Sources/TokamakCore/Views/Text.swift
+++ b/Sources/TokamakCore/Views/Text.swift
@@ -30,10 +30,15 @@
 ///       .italic()
 ///       .underline(true, color: .red)
 public struct Text: View {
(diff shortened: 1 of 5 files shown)


carson-katri:
@MaxDesistov do you know why the operator is not included with the typealias? I don’t understand why that would be...

MaxDesiatov:
IDK, maybe because it's defined in an extension? Although I thought all public members of extensions get exported through typealiases. Maybe I'm missing something, need to have a closer look to say for sure.

PR #179, opened 2020-07-13
Add LazyHGrid and LazyVGrid

carson-katri:
They currently aren't lazy loaded, but once `onAppear` and `onDisappear` use the `IntersectionObserver` API, this should be possible.

### DIFF
--- a/Sources/TokamakDemo/GridDemo.swift
+++ b/Sources/TokamakDemo/GridDemo.swift
@@ -21,7 +21,7 @@ import SwiftUI
 import TokamakDOM
 #endif
 
-@available(OSX 10.16, *)
+@available(OSX 10.16, iOS 14.0, *)
 public struct GridDemo: View {
   public var body: some View {
     Group {
--- a/Sources/TokamakDemo/GridDemo.swift
+++ b/Sources/TokamakDemo/GridDemo.swift
@@ -1,0 +1,56 @@
+// Copyright 2019-2020 Tokamak contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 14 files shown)


Sources/TokamakDemo/TokamakDemo.swift
```diff
@@ -41,23 +41,30 @@ struct TokamakDemoView: View {
...
+          EnvironmentDemo()
+            .font(.system(size: 8))
+        }
+        Group {
+          #if canImport(TokamakDOM)
+          ListDemo().listStyle(InsetGroupedListStyle())
+          #else
+          ListDemo()
+          #endif
+          if #available(OSX 10.16, *) {
```
MaxDesiatov, line 59: Please also make it available on iOS.

```suggestion
          if #available(OSX 10.16, iOS 14.0, *) {
```

Sources/TokamakDemo/TokamakDemo.swift
```diff
@@ -41,23 +41,30 @@ struct TokamakDemoView: View {
...
+            OutlineGroupDemo()
+          }
+          ColorDemo()
+            .padding()
+          if #available(OSX 10.16, *) {
```
MaxDesiatov, line 64:
```suggestion
          if #available(OSX 10.16, iOS 14.0, *) {
```

Sources/TokamakDemo/GridDemo.swift
```diff
@@ -0,0 +1,56 @@
...
+//  Created by Carson Katri on 7/13/20.
+//
+
+#if canImport(SwiftUI)
+import SwiftUI
+#else
+import TokamakDOM
+#endif
+
+@available(OSX 10.16, *)
```
MaxDesiatov, line 24:
```suggestion
@available(OSX 10.16, iOS 14.0, *)
```

PR #192, opened 2020-07-17
Add `TokamakShim` module to simplify imports

MaxDesiatov:
Adding this module as a dependency, Tokamak users would only need to add a single import regardless of the platform they're targeting. Thus, instead of

```swift
#if canImport(SwiftUI)
import SwiftUI
#else
import TokamakDOM
#endif
```

a single `import TokamakShim` is enough. `TokamakShim` re-exports correct modules based on a target platform.

I've also renamed the `TokamakDemo Native` directory to `NativeDemo` for brevity.

`xcodebuild` output in the `macos_demo_build` job is now passed to `xcpretty` for more readable build logs.


Package.swift
```diff
@@ -20,6 +21,10 @@ let package = Package(
       name: "TokamakDOM",
       targets: ["TokamakDOM"]
     ),
+    .library(
+      name: "TokamakShim",
+      targets: ["TokamakShim"]
+    ),
```
j-f1, line 27: This is awesome! Why not call it just `Tokamak`?

Sources/TokamakDOM/Views/HTML.swift
```diff
@@ -24,6 +24,7 @@ public typealias View = TokamakCore.View
 public typealias AnyView = TokamakCore.AnyView
 public typealias EmptyView = TokamakCore.EmptyView
 public typealias State = TokamakCore.State
+public typealias Binding = TokamakCore.Binding
 public typealias ObservableObject = TokamakCore.ObservableObject
 public typealias Published = TokamakCore.Published
 public typealias ObservedObject = TokamakCore.ObservedObject
```
j-f1, line 30: A little bit off-topic for this PR but it might make sense to re-export these things in a special file. Maybe `TokamakDOM/CoreExports.swift`? I can file an issue/make a PR if it makes sense.

MaxDesiatov: Yes, it does make sense, issues/PRs appreciated 👍

Package.swift
MaxDesiatov: It could be named just `Tokamak`, I'm only worried that it would hide the actual purpose of the module, which does re-exporting based on the platform. This exporting logic may not work for all cases though. As soon as we have multiple renderers, I could see some people using `TokamakDOM` together with `TokamakHTML` for pre-rendering for example. Or an iOS/macOS app that uses `TokamakHTML` for rendering static websites. In the latter cases you'd import both `SwiftUI` and `TokamakHTML`. I wonder if users would expect an HTML renderer if they import just `Tokamak` on iOS/macOS, while `TokamakShim` points to the platform-per-renderer switching more clearly in my opinion.

IDK, this is just my logic, which may not reflect the actual reality. What do you think @j-f1 @carson-katri?

carson-katri: I agree with that logic.

Sources/TokamakDOM/Core.swift
```diff
@@ -26,6 +26,7 @@ public typealias ObservableObject = TokamakCore.ObservableObject
 public typealias ObservedObject = TokamakCore.ObservedObject
 public typealias Published = TokamakCore.Published
 public typealias State = TokamakCore.State
+public typealias Binding = TokamakCore.Binding
```
j-f1, line 29: Can you alphabetize?

MaxDesiatov: Sure, done 👍

PR #195, opened 2020-07-20
Add SwiftUI App Lifecycle

carson-katri:
This PR adds the `App` protocol, as well as `Scene`s, modifiers, and changes to the renderer to make building apps entirely with Tokamak simpler:

```swift
struct MyApp: App {
  var body: some Scene {
    WindowGroup("Tokamak") {
      Text("Hello, world!")
    }
  }
}

MyApp.main()
```

I've also implemented `@AppStorage` and `@SceneStorage` with the ability for renderers to override the default containers:

```swift
@AppStorage("count") var count: Int = 0 // Persists/updates across tabs.
@SceneStorage("count") var sceneCount: Int = 0  // Persists across single-tab reloads.
```

`AppStorage` will automatically update all tabs when one changes the value. `SceneStorage` will automatically be loaded on page-load as well as when any `View` in the app updates it. `TokamakDOM` uses [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) and [`sessionStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage).

This PR also contains the `ScenePhase` environment value:
```swift
@Environment(\.scenePhase) var phase: ScenePhase
```
* `.active`: Tab is active in the window of the browser.
* `.background`: Tab is open, but not active.

This automatically updates with the [`PageVisibility API`](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API).

PR #202, opened 2020-07-22
Generalize environment injection for all mounted elements

MaxDesiatov:
This is a suggestion for #195 that reduces code duplication.


carson-katri:
Thanks for this!

j-f1:
I don’t know much about this area so I’m leaving it up to y’all

PR #206, opened 2020-07-23
Fix typo in `TokamakDemo.swift`

MaxDesiatov:


MaxDesiatov:
No problem! I myself probably can't, so watch out 😄

PR #207, opened 2020-07-23
Warn against beta versions of Xcode in README.md

MaxDesiatov:
Seems like Xcode 12 beta 3 is particularly problematic due to https://github.com/swiftwasm/carton/issues/65

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,10 @@ app.
 
 ## Requirements for app developers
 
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
Yes, who knows what beta 4 will bring and if we'll manage to fix the beta 3 bug on our side by that point. I'd rather not spend time on enumerating specific beta build numbers that were verified to work

PR #210, opened 2020-07-24
Implement `SidebarListStyle`

Outcue:
First pass at the SidebarListStyle.

For this to be truly complete, work needs to be completed on ButtonStyles. I can look at that next. NavigationLinks in a Sidebar are Text items that draw a blue background when clicked (and perhaps when item is active.)

### DIFF
--- a/Sources/TokamakCore/Views/Containers/List.swift
+++ b/Sources/TokamakCore/Views/Containers/List.swift
@@ -120,7 +120,7 @@ public struct _ListRow {
   public static func listRow<V: View>(_ view: V, _ style: ListStyle, isLast: Bool) -> some View {
     (style as? ListStyleDeferredToRenderer)?.listRow(view) ??
       AnyView(view.padding([.trailing, .top, .bottom]))
-    if !isLast && style as? SidebarListStyle == nil {
+    if !isLast && style.hasDividers {
       Divider()
     }
   }
--- a/Sources/TokamakCore/Views/Containers/List.swift
+++ b/Sources/TokamakCore/Views/Containers/List.swift
@@ -120,7 +120,7 @@ public struct _ListRow {
   public static func listRow<V: View>(_ view: V, _ style: ListStyle, isLast: Bool) -> some View {
     (style as? ListStyleDeferredToRenderer)?.listRow(view) ??
       AnyView(view.padding([.trailing, .top, .bottom]))
-    if !isLast {
+    if !isLast && style as? SidebarListStyle == nil {
       Divider()
     }
   }
--- a/Sources/TokamakDemo/TokamakDemo.swift
+++ b/Sources/TokamakDemo/TokamakDemo.swift
@@ -131,6 +131,8 @@ struct TokamakDemoView: View {
           } else {
             #if os(WASI)
             Text(link.id)
+            #elseif os(macOS)
+            Text(link.id).opacity(0.5)
             #else
             HStack {
               Text(link.id)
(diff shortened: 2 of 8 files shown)


Outcue:
I couldn't figure out how to add the label in order to pass the check-labels check.

j-f1:
Can you also remove the dividers?

Outcue:
> Can you also remove the dividers?

Sure. Let me take a look.

Outcue:
I added a check for usage of the SidebarListStyle. This isn't my favorite way of checking capabilities, but I was unsure if I should add some sort of private protocol for "showsDividers" or not.

j-f1:
I think that's a good way for now and we can re-evaluate later on if needed.

Sources/TokamakDemo/TokamakDemo.swift
```diff
@@ -135,7 +146,7 @@ struct TokamakDemoView: View {
         title: "Demos"
       )
       #if os(WASI)
-      return AnyView(list)
+      return AnyView(list.listStyle(SidebarListStyle()))
```
j-f1, line 149: You should be able to merge this in like I did in the `sidebarDemo`

Outcue: I am not sure what you want me to return here. In the change you made, you returned a NavigationItem. In this case, we just are returning a boring old list if the SidebarListStyle is not available.

j-f1:
```swift
if #available(iOS 14.0, *) {
  return AnyView(list.listStyle(SidebarListStyle()))
} else {
  return AnyView(list)
}
```

Outcue: Isn't that exactly what is on lines 150-156 here?

j-f1: Yep! The `if os(WASI)` part shouldn’t be necessary because the `*` in the `#available` includes the WASI platform.

Outcue: Got it.

Sources/TokamakCore/Views/Containers/List.swift
```diff
@@ -120,7 +120,7 @@ public struct _ListRow {
   public static func listRow<V: View>(_ view: V, _ style: ListStyle, isLast: Bool) -> some View {
     (style as? ListStyleDeferredToRenderer)?.listRow(view) ??
       AnyView(view.padding([.trailing, .top, .bottom]))
-    if !isLast {
+    if !isLast && style as? SidebarListStyle == nil {
```
carson-katri, line 123: You should probably just add a `var addDividers: Bool { get }` to the `ListStyleDeferredToRenderer` protocol, then provide a default value of `true`.

Outcue: I added "hasDividers" to the ListStyle. It wasn't clear how to add it to ListStyleDeferredToRenderer and have the code be clean. 

Here is where the property is checked:

```
@ViewBuilder
  public static func listRow<V: View>(_ view: V, _ style: ListStyle, isLast: Bool) -> some View {
    (style as? ListStyleDeferredToRenderer)?.listRow(view) ??
      AnyView(view.padding([.trailing, .top, .bottom]))
    if !isLast && style.hasDividers {
      Divider()
    }
  }
}
```

MaxDesiatov:
Thank you @Outcue!

PR #213, opened 2020-07-25
Implement `DynamicProperty`

carson-katri:
This matches the SwiftUI protocol `DynamicProperty`. I also modified `State` to use OpenCombine, let me know if you want that changed back. I also fixed an issue with environment injection in the process.


Sources/TokamakCore/MountedViews/MountedCompositeElement.swift
```diff
@@ -30,7 +30,7 @@ class MountedCompositeElement<R: Renderer>: MountedElement<R>, Hashable {
   var mountedChildren = [MountedElement<R>]()
   let parentTarget: R.TargetType
 
-  var state = [Any]()
+  var state = [StateLocation]()
```
MaxDesiatov, line 33: I'm a bit unsure about the overhead of the new `StateLocation` type. While previously the value was stored directly (boxed as `Any` though), now there's an associated `CurrentValueSubject` object, so at least from the memory perspective I wonder if it consumes a few more bytes per state object (at least in theory). In terms of performance I'm not able to say anything until we have a proper benchmark suite. And we totally should have benchmarks sooner rather than later, I'm chipping away small bits slowly in that direction to have a proper test suite first.

Sources/TokamakCore/State/State.swift
```diff
@@ -15,40 +15,45 @@
...
       fatalError("\(#function) not available outside of `body`")
     }
-    // swiftlint:disable:next force_cast
-    return .init(get: { getter() as! Value }, set: { setter($0) })
+    return .init {
+      // swiftlint:disable:next force_cast
+      _location.value as! Value
+    } set: {
+      _location.value = $0
+    }
```
MaxDesiatov, line 52: I hope we could avoid the multiple trailing closure syntax in cases such as this, where closures are symmetrical, but their labels no longer are.

Sources/TokamakCore/MountedViews/MountedCompositeElement.swift
carson-katri: Makes sense, I'll revert it.

PR #215, opened 2020-07-26
Generalize style and environment in `DOMRenderer`

MaxDesiatov:
Moves `DOMNode` to a separate file and reduces style and environment duplication in overloaded `DOMRenderer` initializers.

### DIFF
--- a/Sources/TokamakDOM/DOMRenderer.swift
+++ b/Sources/TokamakDOM/DOMRenderer.swift
@@ -30,7 +30,10 @@ extension EnvironmentValues {
   }
 }
 
-// FIXME: doesn't this belong to TokamakCore?
+/** `SpacerContainer` is part of TokamakDOM, as not all renderers will handle flexible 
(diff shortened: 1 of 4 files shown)


Sources/TokamakDOM/DOMRenderer.swift
```diff
@@ -18,43 +18,19 @@
...
-        return .undefined
-      }
-      _ = ref.addEventListener!(event, jsClosure)
-      self.listeners[event] = jsClosure
-    }
+    return environment
   }
 }
 
+// FIXME: doesn't this belong to TokamakCore?
```
carson-katri, line 33: `SpacerContainer` is part of TokamakDOM, as not all renderers will handle flexible sizing the way browsers do. Their parent element could already know that if a child is requesting full width, then it needs to expand.

Although, it also makes sense to move this to TokamakCore so that any Renderer that does require this information doesn't need to do so from scratch...

PR #217, opened 2020-07-27
Add `DefaultApp` type to simplify `DOMRenderer.init`

MaxDesiatov:
This means that the `View` initializer of `DOMRenderer` now can delegate to the `App` initializer by creating a wrapping `DefaultApp`. It would simplify https://github.com/swiftwasm/Tokamak/pull/136 for me, where I no longer need to implement color scheme observation for these two different codepaths separately.

### DIFF
--- a/Sources/TokamakDOM/App.swift
+++ b/Sources/TokamakDOM/App.swift
@@ -23,7 +23,6 @@ private enum ScenePhaseObserver {
   static var publisher = CurrentValueSubject<ScenePhase, Never>(.active)
 
   static func observe() {
-    let document = JSObjectRef.global.document.object!
     _ = document.addEventListener!("visibilitychange", JSClosure { _ in
       let visibilityState = document.visibilityState.string
       if visibilityState == "visible" {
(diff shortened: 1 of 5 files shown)


carson-katri:
The only concern I have is that the body tag will have its styles overridden when you’re just rendering a View. Before this wasn’t the case, so you could embed tokamak in an existing project w/o it affecting any of your existing styles.

Maybe add a flag to check in launch if the body styles should be added or not?

Sources/TokamakDOM/App.swift
```diff
@@ -44,12 +43,11 @@ extension App {
...
-                             _ rootEnvironment: EnvironmentValues) {
-    let document = JSObjectRef.global.document.object!
-    let body = document.body.object!
-    let head = document.head.object!
-    body.style = "margin: 0;"
+  public static func _launch(_ app: Self, _ rootEnvironment: EnvironmentValues) {
+    let body = TokamakDOM.body
+    if body.style == .undefined {
+      body.style = "margin: 0;"
+    }
```
MaxDesiatov, line 50: @carson-katri what do you think about this check?

carson-katri: That should do it 👍

PR #223, opened 2020-07-28
Fix environment changes causing remounted scenes with lost state

### LINKED ISSUE #222: Runtime issues with dynamic properties in `App` types
Here's a small example:

```swift
struct TokamakDemoApp: App {
  @ObservedObject var count = Count(value: 0)

  var body: some Scene {
    WindowGroup("Tokamak Demo") {
        Text("\(count.value)")
    }
  }
}
```

crashes with:

```
RuntimeError: memory access out of bounds
    at strlen (<anonymous>:wasm-function[47490]:0xb8643d)
    at $sSS7cStringSSSPys4Int8VG_tcfCTf4nd_n (<anonymous>:wasm-function[20390]:0x556bae)
    at $sSS7cStringSSSPys4Int8VG_tcfC (<anonymous>:wasm-function[20389]:0x556b68)
    at $s7Runtime19NominalMetadataTypePAAE11mangledNameSSyF (<anonymous>:wasm-function[8313]:0x2150cb)
    at $s7Runtime13ClassMetadataV10toTypeInfoAA0eF0VyF (<anonymous>:wasm-function[8168]:0x20e6a9)
    at $s7Runtime13ClassMetadataVAA19TypeInfoConvertibleA2aDP02todE0AA0dE0VyFTW (<anonymous>:wasm-function[8192]:0x20f2e3)
    at $s7Runtime8typeInfo2ofAA04TypeC0VypXp_tKF (<anonymous>:wasm-function[8660]:0x223bcd)
    at $s11OpenCombine16ObservableObjectPA2A0cD9PublisherC0d10WillChangeE0RtzrlE06objectfG0AEvg (<anonymous>:wasm-function[1705]:0x88986)
    at $s11TokamakDemo5CountC11OpenCombine16ObservableObjectAadEP16objectWillChange0giJ9PublisherQzvgTW (<anonymous>:wasm-function[16762]:0x443ce3)
    at $s11TokamakCore14ObservedObjectV16objectWillChange11OpenCombine12AnyPublisherVyyts5NeverOGvg (<anonymous>:wasm-function[11415]:0x2b6422)
```

This occurs in Views as well.

*Update:*

`@State` also has some issues. Here's a small example:
```swift
struct TokamakDemoApp: App {
  @State private var count: Int = 0
  var body: some Scene {
    WindowGroup("Tokamak Demo") {
      Button("\(count)") { count += 1 }
    }
  }
}
```
The `count` starts with the number `207884`. 🤨

MaxDesiatov:
Implements support for custom scenes by adding the new `MountedScene` type and fixes environment propagation from mounted parent to mounted child elements.

This will unblock https://github.com/swiftwasm/Tokamak/pull/136 and potentially https://github.com/swiftwasm/Tokamak/pull/214. In the former scenes are always completely unmounted and remounted from scratch when root environment changes, which causes descendants to lose all `@State` values. I saw some environment propagation issues in the latter, but not sure if those were caused by the lack of correct scene updates, just wanted to tackle the usual suspects first.

I've also improved reconciler-related doc comments to clarify some of the design desicions and naming.

Resolves #222.

### DIFF
--- a/Sources/TokamakCore/App/Scenes/Scene.swift
+++ b/Sources/TokamakCore/App/Scenes/Scene.swift
@@ -27,10 +27,6 @@ protocol TitledScene {
   var title: Text? { get }
 }
 
-protocol ViewContainingScene {
-  var anyContent: AnyView { get }
-}
-
 protocol ParentScene {
   var children: [_AnyScene] { get }
 }
--- a/Sources/TokamakCore/MountedViews/MountedApp.swift
+++ b/Sources/TokamakCore/MountedViews/MountedApp.swift
@@ -34,57 +34,25 @@ final class MountedApp<R: Renderer>: MountedCompositeElement<R> {
     mountedChildren.forEach { $0.unmount(with: reconciler) }
   }
 
-  func mountChild<S: Scene>(_ childBody: S) -> MountedElement<R> {
+  private func mountChild<S: Scene>(_ childBody: S) -> MountedElement<R> {
     let mountedScene: MountedScene<R> = childBody.makeMountedView(parentTarget,
                                                                   environmentValues)
     if let title = mountedScene.title {
       // swiftlint:disable force_cast
-      (app.appType as! _TitledApp.Type)._setTitle(title)
+      (app.type as! _TitledApp.Type)._setTitle(title)
     }
     return mountedScene.body
   }
 
   override func update(with reconciler: StackReconciler<R>) {
-    // FIXME: for now without properly handling `Group` mounted composite views have only
-    // a single element in `mountedChildren`, but this will change when
-    // fragments are implemented and this switch should be rewritten to compare
-    // all elements in `mountedChildren`
-
-    // swiftlint:disable:next force_try
-    let appInfo = try! typeInfo(of: app.appType)
-    appInfo.injectEnvironment(from: environmentValues, into: &app.app)
-
-    switch (mountedChildren.last, reconciler.render(mountedApp: self)) {
-    // no mounted children, but children available now
-    case let (nil, childBody):
-      let child: MountedElement<R> = mountChild(childBody)
-      mountedChildren = [child]
-      child.mount(with: reconciler)
-
-    // some mounted children
-    case let (wrapper?, childBody):
-      let childBodyType = (childBody as? AnyView)?.type ?? type(of: childBody)
-
-      // FIXME: no idea if using `mangledName` is reliable, but seems to be the only way to get
-      // a name of a type constructor in runtime. Should definitely check if these are different
-      // across modules, otherwise can cause problems with views with same names in different
-      // modules.
-
-      // new child has the same type as existing child
-      // swiftlint:disable:next force_try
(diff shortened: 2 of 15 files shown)


Sources/TokamakCore/App/Scenes/Scene.swift
```diff
@@ -27,10 +27,6 @@ protocol TitledScene {
   var title: Text? { get }
 }
 
-protocol ViewContainingScene {
-  var anyContent: AnyView { get }
-}
-
```
MaxDesiatov, line 33: There doesn't seem to be a difference between `ViewContainingScene` and `SceneDeferredToRenderer`, so I'm only keeping the latter.

Sources/TokamakCore/MountedViews/MountedApp.swift
```diff
@@ -34,49 +34,43 @@ final class MountedApp<R: Renderer>: MountedCompositeElement<R> {
     mountedChildren.forEach { $0.unmount(with: reconciler) }
   }
 
-  private func mountChild<S: Scene>(_ childBody: S) -> MountedElement<R> {
```
MaxDesiatov, line 37: In practice `S` is always `_AnyScene` here as everything is type-erased by this point, so no need for it to be generic.

Sources/TokamakCore/MountedViews/MountedApp.swift
```diff
@@ -34,49 +34,43 @@ final class MountedApp<R: Renderer>: MountedCompositeElement<R> {
...
   override func update(with reconciler: StackReconciler<R>) {
     let element = reconciler.render(mountedApp: self)
     reconciler.reconcile(
       self,
       with: element,
-      getElementType: { ($0 as? _AnyScene)?.type ?? type(of: $0) },
-      updateChild: { $0.scene = _AnyScene(element) },
+      getElementType: { $0.type },
+      updateChild: {
+        $0.environmentValues = environmentValues
```
MaxDesiatov, line 53: This is where environment was not passed from a parent to its children previously.

Sources/TokamakCore/MountedViews/MountedCompositeElement.swift
```diff
@@ -27,28 +27,23 @@ class MountedCompositeElement<R: Renderer>: MountedElement<R>, Hashable {
     hasher.combine(ObjectIdentifier(self))
   }
 
-  var mountedChildren = [MountedElement<R>]()
   let parentTarget: R.TargetType
 
   var state = [Any]()
   var subscriptions = [AnyCancellable]()
-  var environmentValues: EnvironmentValues
```
MaxDesiatov, line 35: Both `environmentValues` and `mountedChildren` now live in the base `MountedElement` superclass as there are no subclasses that don't need them except the `MountedEmptyView`. I think the overhead of `MountedEmptyView` having empty children and environment containers on it is justified for now when considering the simplification we get here.

MaxDesiatov:
No problem, thanks for the review!

PR #225, opened 2020-07-29
Remove `_AnyApp`, add `Renderer.AppType` associated type

### LINKED ISSUE #224: Establish App lifecycle as the only way to start rendering
The reasoning here is that the only way to start rendering should be through creating an `App` type and calling `main()` on it. This would allow us to remove the `View`-based `DOMRenderer` initializer and also thread the type of `App` through `DOMRenderer` and `StackReconciler` through generics without erasing it with `_AnyApp`. Then potentially we could remove `_AnyApp` and the indirection it requires altogether. `_AnyApp` is not present in the upstream API anyway 🤔

What do you think @carson-katri @j-f1?

MaxDesiatov:
Closes #224

This introduces a crash in the scene environment code and I'm still not sure why all known crashes are gone in https://github.com/swiftwasm/Tokamak/pull/223. Keeping this as a draft until we get more clarity on how to resolve this.


Sources/TokamakCore/MountedViews/MountedApp.swift
```diff
@@ -48,7 +47,7 @@ final class MountedApp<R: Renderer>: MountedCompositeElement<R> {
     reconciler.reconcile(
       self,
       with: element,
-      getElementType: { $0.type },
+      getElementType: { type(of: $0) },
```
j-f1, line 50: Could this be?

```suggestion
      getElementType: type(of:),
```

Still a bit of a swift beginner

MaxDesiatov:
Closing this as outdated. The original crash was resolved, and this branch has diverged from `main` quite significantly for this refactor to have a benefit on its own.

PR #227, opened 2020-07-29
Remove `DefaultApp`, make `DOMRenderer` internal

### LINKED ISSUE #224: Establish App lifecycle as the only way to start rendering

MaxDesiatov:
Removes the `View`-based initializer of `DOMRenderer` which no longer leaves any `public` initializers on it, means we can make it fully internal. `DOMNode` is now internal too, which is great as it was an implementation detail anyway. Corollary, `DefaultApp` is no longer needed.

`Target` was cleaned up is it doesn't need to hold `App` or `Scene` values, now it's just a simple protocol.

I've updated `README.md` to show usage of the `App` protocol in the basic example.

Closes #224.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,23 +48,18 @@ struct Counter: View {
     }
   }
 }
-```
 
-You can then render your view in any DOM node captured with
-[JavaScriptKit](https://github.com/kateinoigakukun/JavaScriptKit/), just
-pass it as an argument to the `DOMRenderer` initializer together with your view:
+struct CounterApp: App {
+  var body: some Scene {
+    WindowGroup("Counter Demo") {
+      Counter(count: 5, limit: 15)
+    }
+  }
+}
 
-```swift
-import JavaScriptKit
-import TokamakDOM
-
-let document = JSObjectRef.global.document.object!
-
-let divElement = document.createElement!("div").object!
-let renderer = DOMRenderer(Counter(count: 5, limit: 15), divElement)
-
-let body = document.body.object!
-_ = body.appendChild!(divElement)
+// @main attribute is not supported in SwiftPM apps.
+// See https://bugs.swift.org/browse/SR-12683 for more details.
+CounterApp.main()
 ```
 
 ### Arbitrary HTML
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -108,9 +103,9 @@ app.
 ## Requirements for app developers
 
 - macOS 10.15 and Xcode 11.4/11.5/11.6 for macOS. Xcode betas are currently not supported. You can have
-those installed, but please make sure you use 
-[`xcode-select`](https://developer.apple.com/library/archive/technotes/tn2339/_index.html#//apple_ref/doc/uid/DTS40014588-CH1-HOW_DO_I_SELECT_THE_DEFAULT_VERSION_OF_XCODE_TO_USE_FOR_MY_COMMAND_LINE_TOOLS_) 
-to point it to a release version of Xcode.
+  those installed, but please make sure you use
+  [`xcode-select`](https://developer.apple.com/library/archive/technotes/tn2339/_index.html#//apple_ref/doc/uid/DTS40014588-CH1-HOW_DO_I_SELECT_THE_DEFAULT_VERSION_OF_XCODE_TO_USE_FOR_MY_COMMAND_LINE_TOOLS_)
+  to point it to a release version of Xcode.
 - [Swift 5.2 or later](https://swift.org/download/) for Linux.
```
carson-katri, line 109: Do we need Swift 5.3 for the conditions inside the `Package.swift`?

MaxDesiatov: We do, but that's the SwiftWasm toolchain version, the requirements in `README.md` are just to install `carton`. `carton` itself will pull the recommended toolchain/SDK version.

carson-katri: Got it, thanks.

PR #228, opened 2020-07-30
Add `Slider` implementation

j-f1:
Status:

- [x] Slider visible
- [x] Updates `value` binding
- [x] Supports `bounds`
- [x] Supports `step`
- [ ] Marks steps with ticks (in SwiftUI, this only happens on macOS)
- [ ] Displays the `label` (in SwiftUI, this only happens on macOS)
- [x] reports `onEditingChanged`
- [x] allows passing in any `BinaryFloatingPoint` (how do I do this? I can’t store one directly)
- [x] thorough demo
- [ ] slider expands to fill available space


Sources/TokamakCore/Views/Selectors/Slider.swift
```diff
@@ -0,0 +1,164 @@
...
+///
+/// Available when `Label` and `ValueLabel` conform to `View`.
+public struct Slider<Label, ValueLabel>: View where Label: View, ValueLabel: View {
+  let label: Label
+  let minValueLabel: ValueLabel
+  let maxValueLabel: ValueLabel
+  let valueBinding: Binding<Double>
+  let bounds: ClosedRange<Double>
+  let step: _SliderStep // <Double>
+  // “For example, on iOS, a Slider is considered to be actively editing while the user is touching the knob and sliding it around the track.”
```
ie-ahm-robox, line 32: - :no_entry_sign: Line should be 100 characters or less: currently 142 characters (`line_length`)

ie-ahm-robox:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  **Sources/TokamakDemo/SliderDemo.swift#L22** - Line should be 100 characters or less: currently 115 characters (`line_length`)
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against 8b99d5bbd9ec376e9fe109d1825be846b0f7e1fc
</p>

Sources/TokamakDemo/SliderDemo.swift
```diff
@@ -0,0 +1,28 @@
...
+// limitations under the License.
+
+import TokamakShim
+
+struct SliderDemo: View {
+  @State private var value: Double = 0
+
+  var body: some View {
+    VStack {
+      Slider(value: $value, in: 0...1, step: 0.1, minimumValueLabel: Text("min"), maximumValueLabel: Text("max")) {
```
ie-ahm-robox, line 22: - :no_entry_sign: Line should be 100 characters or less: currently 115 characters (`line_length`)

carson-katri:
For #8 you can make your initializer generic with the `BinaryFloatingPoint` constraint (instead of the `typealias V`), then create a `Binding` that casts the get/set of the passed in `Binding<V>` to Double in the init body.

j-f1:
ok this whole thing is really buggy in native and I’m wondering if it’s better on 10.16.

- when you click on the slider in the demo it stays in the “editing” state because it immediately calls `onEditingChanged` with first `false` then `true`
- if you change `step: 0.1` to `step: 0.3` the slider position snaps appropriately but the ticks are wrong
  Screenshot_2020-07-31 16 58 48

MaxDesiatov:
IDK how thoroughly we'd like to follow the macOS behavior. Maybe if the iOS implementation is not as detailed, and the browser version is closer to iOS than macOS, we could stick to that as a first approximation?

j-f1:
OK the thing that’s limiting the slider width is that the slider is wrapped in an `HStack` to allow showing min/max views. However, the `HStack` doesn’t want to expand to fill the space horizontally. Should I add `width: 100%` to all `HStack`s to enable this behavior?

MaxDesiatov:
I previously tried to add `width: 100%` to all `HStack`s to make destination views sizing of `NavigationView` more consistent with SwiftUI, but it broke other things, probably scrolling. I hope there's still a way to resolve that, maybe separately from this PR?

MaxDesiatov:
BTW, `macos_build` is failing due to my merge, I'm preparing a PR to fix that.

j-f1:
I think I’m done with this PR, and I plan to open an issue for the expand-to-fill-the-width thing.

Sources/TokamakCore/Views/Selectors/Slider.swift
```diff
@@ -0,0 +1,169 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+public enum _SliderStep {
+  case any
+  case discrete(Double.Stride)
+}
+
+func convert<T>(_ binding: Binding<T>) -> Binding<Double> where T: BinaryFloatingPoint {
```
carson-katri, line 20: I think these `convert` functions should either be static methods on `Slider`, or changed to `fileprivate`.

Sources/TokamakCore/Views/Selectors/Slider.swift
```diff
@@ -0,0 +1,169 @@
...
+    valueBinding = convert(value)
+    self.bounds = convert(bounds)
+    step = .any
+    self.onEditingChanged = onEditingChanged
+  }
+
+  public init<V>(
+    value: Binding<V>,
+    in bounds: ClosedRange<V>,
+    step: V.Stride = 1,
```
MaxDesiatov, line 65: I'll be able to try this in action later today, but just looking at the code, I'm not sure how this overloaded function is distinguished from the one above when the `step` argument is omitted and the default `1` value is supplied? The types of the rest of the arguments seem to be identical to me, but maybe I'm missing something? Is this how SwiftUI defines this as well?

docs/progress.md
```diff
@@ -46,7 +46,7 @@ Table columns:
 | 🚧  | [Toggle](https://developer.apple.com/documentation/swiftui/toggle)           |     |
 | 🚧  | [Picker](https://developer.apple.com/documentation/swiftui/picker)           |     |
 |     | [DatePicker](https://developer.apple.com/documentation/swiftui/datepicker)   |     |
-|     | [Slider](https://developer.apple.com/documentation/swiftui/slider)           |     |
+| ✅  | [Slider](https://developer.apple.com/documentation/swiftui/slider)           |     |
```
MaxDesiatov, line 49: Shouldn't this be 🚧 with the unresolved FIXME in the comment at `onEditingChanged`?

Sources/TokamakCore/Views/Selectors/Slider.swift
carson-katri: From SwiftUI `.swiftinterface`:
```swift
extension Slider {
  @available(tvOS, unavailable)
  public init<V>(value: SwiftUI.Binding<V>, in bounds: Swift.ClosedRange<V> = 0...1, onEditingChanged: @escaping (Swift.Bool) -> Swift.Void = { _ in }, minimumValueLabel: ValueLabel, maximumValueLabel: ValueLabel, @SwiftUI.ViewBuilder label: () -> Label) where V : Swift.BinaryFloatingPoint, V.Stride : Swift.BinaryFloatingPoint
  public init<V>(value: SwiftUI.Binding<V>, in bounds: Swift.ClosedRange<V>, step: V.Stride = 1, onEditingChanged: @escaping (Swift.Bool) -> Swift.Void = { _ in }, minimumValueLabel: ValueLabel, maximumValueLabel: ValueLabel, @SwiftUI.ViewBuilder label: () -> Label) where V : Swift.BinaryFloatingPoint, V.Stride : Swift.BinaryFloatingPoint
}
```
I agree that it seems ambiguous though.

j-f1: If you specify `step`, you must specify `bounds` (ie `Slider(value: $foo, step: 2)` is invalid). Not sure how it deduplicates, though.

Maybe the reason is that tvOS doesn’t support the continuous version so they have the overload so a `Slider(value: $foo, in: 0...100)` for example works on both platforms?

Sources/TokamakDOM/Views/Selectors/Slider.swift
```diff
@@ -0,0 +1,61 @@
...
+      "max": String(proxy.bounds.upperBound),
+      "step": step,
+      "value": String(proxy.valueBinding.wrappedValue),
+      "style": """
+        display: block;
+        width: 100%;
+      """,
+    ]
+
+    // FIXME: The `label` property is never displayed. This might be OK since iOS also doesn’t display the label, but we should make this decision explicit. macOS puts it next to the minValueLabel.
```
ie-ahm-robox, line 40: - :no_entry_sign: Line should be 100 characters or less: currently 197 characters (`line_length`)

Sources/TokamakDOM/Views/Selectors/Slider.swift
```diff
@@ -0,0 +1,61 @@
...
+      HStack {
+        proxy.minValueLabel
+        DynamicHTML(
+          "input",
+          attributes,
+          listeners: [
+            "input": { event in
+              proxy.valueBinding.wrappedValue = Double(event.target.object!.value.string!)!
+            },
+            // FIXME: This does not handle keyboard input. Maybe it should set editing based on focus? Need to check against native platforms.
```
ie-ahm-robox, line 51: - :no_entry_sign: Line should be 100 characters or less: currently 142 characters (`line_length`)

MaxDesiatov (commented):
It seems there's very little contrast in sliders when dark mode is enabled:

Screenshot 2020-08-26 at 11 26 11

MaxDesiatov (approved):
Totally fine to fix the dark mode contrast in a separate PR, I've created #267 to keep track.

PR #229, opened 2020-07-30
Make reconciler tests build and run on macOS

MaxDesiatov:
We can't run our basic reconciler tests in a WASI environment yet because `XCTestExpectation` is not available on WASI as it relies on the presence of `Dispatch`. We can run these tests on macOS though, and even on Linux in the future when Swift 5.3 is available for Linux on GitHub Actions.

My current OpenCombine fork doesn't build on macOS and it was much easier to add a new `CombineShim` module that uses native Combine there.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -39,8 +39,16 @@ let package = Package(
     // Targets can depend on other targets in this package, and on products
     // in packages which this package depends on.
     .target(
+      name: "CombineShim",
+      dependencies: [.product(
+        name: "OpenCombine",
+        package: "OpenCombine",
+        condition: .when(platforms: [.wasi, .linux])
+      )]
+    ),
+    .target(
       name: "TokamakCore",
-      dependencies: ["OpenCombine", "Runtime"]
+      dependencies: ["CombineShim", "Runtime"]
     ),
     .target(
       name: "TokamakDemo",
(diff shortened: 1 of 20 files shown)


carson-katri:
I wasn’t able to build the combine fork on Arch Linux either. Maybe use OpenCombine/OpenCombine instead of the fork for Linux? Unless that’s only a problem on Arch, which I don’t know for sure.

MaxDesiatov:
No, I don't think my fork can be built on non-WASI platforms, I had a few things from `COpenCombineHelpers` torn out manually rushing to make it work on WASI 😅

Package.swift
```diff
@@ -38,17 +38,25 @@ let package = Package(
     // a module or a test suite.
     // Targets can depend on other targets in this package, and on products
     // in packages which this package depends on.
+    .target(
+      name: "CombineShim",
+      dependencies: [.product(
+        name: "OpenCombine",
+        package: "OpenCombine",
+        condition: .when(platforms: [.wasi, .linux])
```
carson-katri, line 46: Could you swap out the `OpenCombine` url in `.package` on line 34 when on Linux?

MaxDesiatov: I don't think so, at least after looking at [the proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0273-swiftpm-conditional-target-dependencies.md). I'll try to submit my changes to the upstream OpenCombine repo, I just have another old open PR there with the `zip` implementation that I need to finish first...

MaxDesiatov:
Yeah, that builds and runs on macOS though unless you pass the `--triple wasm32-unknown-wasi` option to it. This is what `carton test` does automatically, but it won't build because of `XCTestExpectation` unavailable on WASI.

PR #232, opened 2020-07-31
Add `redacted` modifier

carson-katri:
This adds `redacted(reason:)` and `RedactionReason`. Currently, only `Text` is redacted, and this is done by covering the text with a grey `::after` element.


MaxDesiatov:
I guess `macos_build` is failing because GitHub Actions is still on Xcode 12 beta 2

PR #238, opened 2020-08-01
Temporily disable Native Demo build

MaxDesiatov:
Only until Xcode 12 beta 3 is available on GHA.


MaxDesiatov:
`swift test` is still failing as it attempts to build `TokamakDemo` against SwiftUI, which fails because `redacted`/`unredacted` is unavailable. Blocked by https://github.com/actions/virtual-environments/issues/1166

MaxDesiatov:
Conveniently enough, looks like Beta 3 is on GHA now.

Issue #231, opened 2020-07-31
Easily access DOM/JS from Views


carson-katri:
Just wanted to propose adding a new property wrapper to TokamakDOM called `DOMValue` as well as a `domId` modifier, which would allow a `View` to interoperate with a DOM element attributes, and provide better interoperability with other libraries:

```swift
enum ElementAttribute {
  case value, type, `class`, ...
}

@propertyWrapper
struct DOMValue<Value>: DynamicProperty {
  init(_ attribute: ElementAttribute,
       from selector: String,
       transform: (String) -> Value)
  var wrappedValue: Value
  var projectedValue: Binding<Value>
}
extension DOMValue where Value == String {
  init(_ attribute: ElementAttribute,
       from selector: String)
}
extension DOMValue where Value == Int { ... }
...
```

This could use the `MutationObserver` API to update the view when the element changes.

In conjunction with a `.domId` modifier, this would allow better manipulation of elements than currently possible.

```swift
struct ContentView: View {
  @DOMValue(.type, from: "#myButton") var buttonType

  var body: some View {
    Button("Click me!") { ... }
      .domId("myButton")
      .onAppear { buttonType = "submit" }
  }
}
```

j-f1:
What about a React-style API where you have an `@State` (or maybe a custom `@Ref`) binding which you add to a view then pass to `.ref()`?

```swift
struct ContentView: View {
  @Ref var button: JSObjectRef!
  var body: some View {
    Button("Hello, world!") {
      self.button.textContent.string! = "Clicked!"
    }.ref($button)
  }
}
```

carson-katri:
That is good for accessing elements inside Tokamak, but you may also want to access elements from a host app as well... Maybe `@Ref` could also accept a selector String?

MaxDesiatov:
If have something like `ElementAttribute`, it won't work for custom attributes if there are no associated values. Maybe it would have a separate `case custom(String)` to support those? As for DOM access outisde of the Tokamak element tree, would that be substantially different from this?

```swift
struct ContentView: View {
  var body: some View {
    Button("Click me!") { ... }
      .domId("myButton")
      .onAppear { document.querySelector!("#myButton").type = "submit" }
  }
}
```

And then, would we want to have specialized `dom...` modifier for every attribute, or just a plain `domAttributes` to cover all possible attributes with a single modifier?

```swift
struct ContentView: View {
  var body: some View {
    Button("Click me!") { ... }
[...]

PR #239, opened 2020-08-02
Add `GeometryReader` implementation

MaxDesiatov:
This is just an empty API at the moment. I hope it can be implemented purely in the `deferredBody` of `GeometryReader` with [the ResizeObserver API](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) without requiring any tweaks in the `Renderer` protocol or the reconciler.

Seems like I need the `domRef` modifier that writes `JSObjectRef` to a given binding working first, as discussed in #231.


carson-katri:
I can work on implementing the features outlined in that issue

MaxDesiatov:
I started some preliminary work on that, or do you have some code already?

carson-katri:
No, I hadn’t started yet.

MaxDesiatov:
No worries, I'll try to get some draft of that today.

MaxDesiatov:
This is working for me now in isolated example code, but layout bugs exposed by our demo code prevent this from being added to the demo. @carson-katri @j-f1 lmk if you think it would be fine merging this as is and adding a `GeometryReader` demo in a separate PR (smaller PRs is what I prefer personally).

j-f1:
IMO if it's a small bug or something that prevents building it shouldn't be merged in general. But if it's more complex or harder-to-solve bug, it's ok to merge then follow up with an issue or PR.

carson-katri:
What's the layout bug preventing a demo?

MaxDesiatov:
If you launch the demo in this branch in browsers and navigate to the `GeometryReader` item you'll see that the size doesn't change if you change the size of the viewport (I resized the browser window). It does change in the native demo though.

As far as I understand, the destination view of `NavigationView` doesn't fill all available space currently, maybe it's fixed with the `_FlexFrameLayout` changes made in #241?

j-f1:
I got that to work after I added this CSS to `GeometryReader`:

```css
display: flex;
align-items: center;
justify-content: center;
```

…weirdly, the text *isn’t* exactly centered.

MaxDesiatov:
These styles worked only in Safari for me and only for height, width always stays 300. I've pushed these changes for the lack of a better option, will have a look later if #241 could resolve it.

PR #240, opened 2020-08-02
Add `_targetRef` and `_domRef` modifiers

MaxDesiatov:
Resolves partially #231. `_targetRef` is a modifier that can be used by any renderer, while `_domRef` is an adaptation of that for `DOMRenderer`. Both are underscored as they are not available in SwiftUI, and also their stability is currently not so well known to us, we may consider changing this API in the future.

Example use:

```swift
struct DOMRefDemo: View {
  @State var button: JSObjectRef?

  var body: some View {
    Button("Click me") {
      button?.innerHTML = "This text was set directly through a DOM reference"
    }._domRef($button)
  }
}
```

I've also fixed all known line length warnings in this PR.


carson-katri:
Awesome 👍

Sources/TokamakDemo/TokamakDemo.swift
```diff
@@ -125,6 +125,12 @@ var links: [NavItem] {
     gridDemo,
     redactDemo,
   ]
+
+  #if os(WASI)
+  result.append(NavItem("DOM manipulation", destination: DOMRefDemo()))
```
j-f1, line 130: Could you extract this like some of the other demos and replace it with a `NavItem(unavailable: "DOM manipulation")` on native?

MaxDesiatov: Great point!

PR #241, opened 2020-08-03
Add default dark styles for Views

### LINKED ISSUE #237: Add dark color scheme style
`ColorScheme` support is already available in #136, but default styles for dark mode are missing.

carson-katri:
This provides dark mode styles for the background and many elements, such as `Text`, `Button`, `TextField`, etc.

Resolves #237.


Sources/TokamakStaticHTML/Views/Text/Text.swift
```diff
@@ -151,13 +151,13 @@ extension Text: AnyHTML {
...
       \(font == nil ? "font-family: \(Font.Design.default.description);" : "")
-      color: \(color?.description ?? "inherit");
+      color: \((color ?? .primary).cssValue(environment));
       font-style: \(italic ? "italic" : "normal");
       font-weight: \(weight?.value ?? font?._weight.value ?? 400);
       letter-spacing: \(kerning);
       vertical-align: \(baseline == nil ? "baseline" : "\(baseline!)em");
       text-decoration: \(textDecoration);
-      text-decoration-color: \(strikethrough?.1?.description ?? underline?.1?.description
+      text-decoration-color: \(strikethrough?.1?.cssValue(environment) ?? underline?.1?.cssValue(environment)
```
ie-ahm-robox, line 160: - :no_entry_sign: Line should be 100 characters or less: currently 109 characters (`line_length`)

ie-ahm-robox:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  **Sources/TokamakStaticHTML/Shapes/_ShapeView.swift#L42** - Line should be 100 characters or less: currently 104 characters (`line_length`)
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against cdc452b7a5b39e598242e9acae73302eddf75833
</p>

carson-katri:
Let me know if any of the designs/colors need to be tweaked.

Sources/TokamakStaticHTML/Resources/TokamakStyles.swift
```diff
@@ -76,6 +76,34 @@ public let tokamakStyles = """
   height: 1.2em;
   border-radius: .1em;
 }
+
+@media (prefers-color-scheme:dark) {
+  ._tokamak-buttonstyle-default {
+    background-color: rgb(99, 95, 98);
```
j-f1, line 82: Does this work with native-looking buttons in Safari?

Sources/TokamakStaticHTML/Views/Containers/List.swift
```diff
@@ -94,7 +94,12 @@ extension SidebarListStyle: ListStyleDeferredToRenderer {
     AnyView(content
       .padding(.all)
       .padding(.leading, 20)
-      .background(Color(0xF2F2F7))
+      .background(Color._withScheme {
+        switch $0 {
+        case .light: return Color(0xF2F2F7)
+        case .dark: return Color(.sRGB, red: 45 / 255, green: 43 / 255, blue: 48 / 255)
```
j-f1, line 100:
```suggestion
        case .dark: return Color(0x2D2B30)
```

Sources/TokamakStaticHTML/Resources/TokamakStyles.swift
carson-katri: Screen Shot 2020-08-02 at 10 07 42 PM

Overwrites the style, just like Chrome. Is there a better way to do this besides trying to replicate the control?

j-f1: Yeah, you can set `color-scheme: light dark` on the `<html>` element or on the button specifically to get a native dark button in dark mode.

MaxDesiatov:
Chrome and Firefox have these blue underlines for `NavigationLink` items for me (probably should be the same color as text if we display those at all?). In Safari 13.1.2 on Catalina there's no underline, but `NavigationView` destination still has white background. Also, in the `OutlineGroup` demo the disclosure control has very low contrast in dark mode. Placeholder text in the `TextField` demo is also low contrast in Chrome.

Screenshot 2020-08-03 at 14 01 48

Screenshot 2020-08-03 at 14 05 15

Screenshot 2020-08-03 at 14 05 24

carson-katri:
I've already fixed the NavigationLink color, as well as the `DisclosureGroup`. I'm working on a `_NavigationLinkStyle` for the sidebar to make it match the Big Sur styles, with the highlighted background. I'm going to try the `color-scheme: dark;` property to see if that fixes the buttons/textfields.

carson-katri:
Here's the dark scheme Views using the `color-scheme` property.

Screen Shot 2020-08-03 at 11 11 59 AM
Screen Shot 2020-08-03 at 11 13 08 AM
Screen Shot 2020-08-03 at 11 14 25 AM

I think the picker looks weird, but besides that these are much nicer. Except they don't work in Chrome. Is there a way for me to provide a different style for Chrome?

j-f1:
I remember that being an issue in Chrome. I think they’re planning to have custom (non-native-looking) default form controls which would probably support dark mode.

carson-katri:
Ok, I've got the SidebarListStyle rendering the `NavigationLink` elements right. I was also able to switch `NavigationView` to use an `ObservableObject` for the context 🎉 

What are your opinions on Chrome styling. Do we write in default dark mode styles?

Sources/TokamakStaticHTML/Views/Text/Text.swift
```diff
@@ -151,13 +173,13 @@ extension Text: AnyHTML {
...
```
ie-ahm-robox, line 182: - :no_entry_sign: Line should be 100 characters or less: currently 109 characters (`line_length`)

Sources/TokamakStaticHTML/Shapes/_ShapeView.swift
```diff
@@ -37,9 +37,9 @@ extension _ShapeView: ViewDeferredToRenderer {
     if let shapeAttributes = shape as? ShapeAttributes {
       return AnyView(HTML("div", shapeAttributes.attributes(style)) { path })
     } else if let color = style as? Color {
-      return AnyView(HTML("div", ["style": "fill: \(color);"]) { path })
+      return AnyView(HTML("div", ["style": "fill: \(color.cssValue(environment));"]) { path })
     } else if let foregroundColor = foregroundColor {
-      return AnyView(HTML("div", ["style": "fill: \(foregroundColor);"]) { path })
+      return AnyView(HTML("div", ["style": "fill: \(foregroundColor.cssValue(environment));"]) { path })
```
ie-ahm-robox, line 42: - :no_entry_sign: Line should be 100 characters or less: currently 104 characters (`line_length`)

ie-ahm-robox:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  **Sources/TokamakStaticHTML/Views/Text/Text.swift#L127** - Function body should span 50 lines or less excluding comments and whitespace: currently spans 52 lines (`function_body_length`)
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against c0a580653af4009be78f7337387da66b1d26a109
</p>

carson-katri:
Oh, I also had to modify the LayoutModifier to check for an infinite frame. That is:
```swift
.frame(minWidth: 0, maxWidth: .infinity)
.frame(minHeight: 0, maxHeight: .infinity)
```
Will always fill it’s parent on the respective axis.

j-f1:
Great to see you have selected link highlighting working!

I’ve noticed a bug in Safari: when you load the demo, click on a section in a sidebar, switch to a different tab, then come back, the content of the `NavigationView` permanently disappears (the sidebar remains) and clicking other navigation links has no effect.

carson-katri:
Maybe its related to the context now being an `ObservableObject`?

MaxDesiatov:
Were any of you able to reproduce https://github.com/swiftwasm/Tokamak/issues/245 while testing this? I wonder if it's somehow related. Would be in line with the bugs in dark mode support that @j-f1 saw previously.

carson-katri:
No, I haven't been able to reproduce that..

carson-katri:
The `NavigationView` not using the entire size viewport should be fixed. All that's left is to add dark mode styles for `List`.

Sources/TokamakStaticHTML/Views/Text/Text.swift
```diff
@@ -89,22 +89,44 @@ extension Font: StylesConvertible {
...
   public var attributes: [String: String] {
     let proxy = _TextProxy(self)
-    let isRedacted = proxy.redactionReasons.contains(.placeholder)
+    return Self.attributes(from: proxy.modifiers,
+                           environment: proxy.environment)
+  }
+}
+
+extension Text {
+  static func attributes(from modifiers: [_Modifier],
```
ie-ahm-robox, line 127: - :no_entry_sign: Function body should span 50 lines or less excluding comments and whitespace: currently spans 54 lines (`function_body_length`)

carson-katri:
Screen Shot 2020-08-04 at 3 42 42 PM
`NavigationView` and `List` are properly working, partly from to the `_FlexFrameLayout` updates.

Sources/TokamakStaticHTML/Views/Text/Text.swift
ie-ahm-robox, line 127: - :no_entry_sign: Function body should span 50 lines or less excluding comments and whitespace: currently spans 54 lines (`function_body_length`)

j-f1:
I’ve got a weird behavior with `SpacerDemo` in Safari but it might be a rendering artifact: https://youtu.be/Q1wPio2qJGw

carson-katri:
Just tested and I'm getting the same behavior on Big Sur, Chrome seems to be fine.

MaxDesiatov:
Default button style doesn't seem to be working for me in any browser:
Screenshot 2020-08-04 at 22 09 07

In Safari 13.1 on Catalina the destination background is still white:

Screenshot 2020-08-04 at 22 12 24

carson-katri:
The Chrome issue is known, they don’t implement default dark scheme buttons. We could provide our own but then we won’t have the safari built-in style...

I’ll haven’t been able to reproduce the safari bug yet.

MaxDesiatov:
Also, I was able to reproduce the bug where `NavigationView` items list stops working, but only in Firefox. Just switching to some other tab made the destination view disappear and the list stopped working. This has something to do with `ScenePhaseObserver`, because when I comment out `ScenePhaseObserver.observe()` it starts working again.

Sources/TokamakDOM/App/ColorSchemeObserver.swift
```diff
@@ -21,13 +21,23 @@ enum ColorSchemeObserver {
...
     }
     _ = matchMediaDarkScheme.addEventListener!("change", closure)
     Self.closure = closure
+    Self.cancellable = Self.publisher.sink { colorScheme in
+      let systemBackground = { () -> String in
+        switch colorScheme {
+        case .light: return "#FFFFFF"
+        case .dark: return "rgb(38, 38, 38)"
+        }
+      }()
```
MaxDesiatov, line 39: **suggestion:** I don't think there's a need for a closure here:
```suggestion
      let systemBackground: String
      switch colorScheme {
      case .light: systemBackground = "#FFFFFF"
      case .dark: systemBackground = "rgb(38, 38, 38)"
      }
```

Sources/TokamakStaticHTML/Views/Containers/List.swift
```diff
@@ -17,84 +17,219 @@ import TokamakCore
...
 }
 
 extension InsetGroupedListStyle: ListStyleDeferredToRenderer {
   public func listBody<ListBody>(_ content: ListBody) -> AnyView where ListBody: View {
     AnyView(content
-      .padding(.top, 20)
-      .background(Color(0xEEEEEE))
+      .background(Color._withScheme {
+        switch $0 {
+        case .light: return Color(0xEEEEEE)
```
MaxDesiatov, line 138: **suggestion:** Some of these colors are repeated, could we get them as constants, in a separate file maybe?

Sources/TokamakStaticHTML/Views/Navigation/NavigationView.swift
```diff
@@ -22,7 +22,17 @@ extension NavigationView: ViewDeferredToRenderer {
...
     ]) {
-      _NavigationViewProxy(self)
+      _NavigationViewProxy(self).content
+      HTML("div", [
+        "style": """
+        display: flex; flex-direction: column;
+        align-items: center; justify-content: center;
+        flex-grow: 1;
+        height: 100%;
+        """,
```
MaxDesiatov, line 32: **suggestion:** Maybe this can be moved to a separate class, purely so that it's easier on the eyes when going through the DOM tree in browser dev tools.

Sources/TokamakStaticHTML/Views/Text/Text.swift
ie-ahm-robox, line 127: - :no_entry_sign: Function body should span 50 lines or less excluding comments and whitespace: currently spans 54 lines (`function_body_length`)

ie-ahm-robox, line 127: - :no_entry_sign: Function body should span 50 lines or less excluding comments and whitespace: currently spans 54 lines (`function_body_length`)

MaxDesiatov:
>The Chrome issue is known, they don’t implement default dark scheme buttons. We could provide our own but then we won’t have the safari built-in style...

Are you referring to this issue here?

Screenshot 2020-08-04 at 22 09 07

If so, it's reproducible for me in Firefox 79, not just Chrome. In Safari I get just white on white:

Screenshot 2020-08-05 at 10 08 34

carson-katri:
Oh ok. So since Safari is the only browser with actual dark scheme buttons maybe we should just provide our own styles, or for now use black text on the buttons. Lmk which you prefer.

As for Safari, the built in dark scheme button style is pretty much fully transparent. So once the color scheme detection bug is fixed we should be fine there.

MaxDesiatov:
>Oh ok. So since Safari is the only browser with actual dark scheme buttons maybe we should just provide our own styles, or for now use black text on the buttons. Lmk which you prefer.

Wouldn't black text on the buttons be the easiest option requiring much less effort?

j-f1:
Now the text of the button is always black (instead of flipping to white when you mousedown) and clicking on the text no longer activates the button (you have to click on the button’s padding instead.

carson-katri:
Also is it only happening in Safari betas? If so, it may be their bug.

j-f1:
I only see it happening in Safari but I also only happen to have beta versions of Safari installed right now so I can’t check the shipping version. I can try to report it though!

carson-katri:
I tested on Catalina and got the same issue with the buttons.

carson-katri:
Ok, I fixed it by disabling:
```swift
"pointerdown": { _ in isPressed = true },
"pointerup": { _ in isPressed = false },
```
These events are [not supported in Safari](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/pointerdown_event). Is there another way to get this information?

MaxDesiatov:
I guess these could be `mousedown` and `mouseup`, which I initially used before @j-f1 suggested switching to pointer events

MaxDesiatov:
MDN says that the pointer events spec is obsolete, I thought it was the recent one 🤔 https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/pointerdown_event

carson-katri:
`mousedown` and `mouseup` are also not working for me...

j-f1:
Pointer events are [supported in Safari 13](https://caniuse.com/#feat=pointer). It looks like there’s a version 2 of the spec which is why the old one is obsolete.

Sources/TokamakCore/Views/Containers/ForEach.swift
```diff
@@ -80,8 +80,49 @@ public extension ForEach where Data == Range<Int>, ID == Int {
...
+    }
+    set {
+      self[_IDKey.self] = newValue
+    }
+  }
+}
+
+public protocol _AnyIDView {
+  var anyId: AnyHashable { get }
+}
```
MaxDesiatov, line 106: This protocol and the property aren't used anywhere, I also don't see it in the upstream API?

Sources/TokamakStaticHTML/Tokens/BuiltinColors.swift
```diff
@@ -0,0 +1,48 @@
+//
+//  File.swift
+//
```
MaxDesiatov, line 3:
```suggestion
```

Sources/TokamakCore/Views/Containers/ForEach.swift
carson-katri: Ah, sorry. I implemented `IDView` when I was trying different things to store the selection state of the sidebar.

I made `_AnyIDView` so you could check if the View has an ID, and access it. I think it'd be necessary for ForEach/List when animations are added so transitions can be made on the items. I can move ID view to a separate PR if that'd be better?

MaxDesiatov: No worries, keeping it this PR is fine, IMO.

Sources/TokamakStaticHTML/Tokens/BuiltinColors.swift
MaxDesiatov: @carson-katri I think this is the only thing that would be great to fix before merging

j-f1:
Two issues in this branch in Safari that I can’t reproduce on `main`: 

- clicking the counter until it should show “Limit Exceeded” instead throws an error
- clicking the button in the DOM Reference demo does nothing or briefly shows the changed text before reverting

carson-katri:
I can't replicate 1, but fixed 2.

MaxDesiatov:
Same, I wasn't able to replicate 1 on my side

PR #243, opened 2020-08-03
Split demo list into sections

j-f1:


Sources/TokamakDemo/TokamakDemo.swift
```diff
@@ -40,122 +40,108 @@ struct NavItem: Identifiable {
...
+            NavItem("Path", destination: PathDemo())
+            NavItem("Environment", destination: EnvironmentDemo().font(.system(size: 8)))
+            NavItem("Color", destination: ColorDemo())
+            if #available(OSX 11.0, iOS 14.0, *) {
+              NavItem("AppStorage", destination: AppStorageDemo())
+            } else {
+              NavItem(unavailable: "AppStorage")
+            }
+            if #available(OSX 11.0, iOS 14.0, *) {
+              NavItem("Redact", destination: RedactDemo())
```
MaxDesiatov, line 137: I'm still not sure if this should be "Redacted", not "Redact"? But I'm not a native English speaker, so maybe the latter does make more sense in this context?

j-f1: Good point! I went with “Redaction” which I think makes a little more sense.

PR #249, opened 2020-08-05
Fix color scheme observer crashes in Safari

### LINKED ISSUE #245: Dark mode detection causes crashes in Safari
I currently see this as soon as the page loads:

```
[Error] Fatal error: Unexpectedly found nil while unwrapping an Optional value: 
file Sources/TokamakDOM/App/ColorSchemeObserver.swift, line 30
```

Not reproducible in other browsers in my testing

MaxDesiatov:
Resolve #245. Turns out `matchMediaDarkScheme` object doesn't have `addEventListener`, but only `addListener` in Safari 13.1.2. IDK why this wasn't reproduced elsewhere, maybe it was changed in Safari minor version that you don't have yet? `¯\_(ツ)_/¯`

### DIFF
--- a/Sources/TokamakDOM/App/ColorSchemeObserver.swift
+++ b/Sources/TokamakDOM/App/ColorSchemeObserver.swift
@@ -27,7 +27,7 @@ enum ColorSchemeObserver {
       publisher.value = .init(matchMediaDarkScheme: $0[0].object!)
       return .undefined
     }
-    _ = matchMediaDarkScheme.addEventListener!("change", closure)
+    _ = matchMediaDarkScheme.addListener!(closure)
     Self.closure = closure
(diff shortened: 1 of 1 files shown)


j-f1:
It’s probably because I’m using Safari Technology Preview (Release 110, Safari 14.0, WebKit 15610.1.21.0.2).

MaxDesiatov:
Huh, I thought that Safari TP is completely broken because of https://github.com/swiftwasm/carton/issues/25

j-f1:
Yeah, I’m using a local override on `dev.js` to run `delete BigInt;` before everything else.

carson-katri:
I'm also using that override to use Safari on Big Sur.

PR #251, opened 2020-08-05
Link to the renderers guide from `README.md`

MaxDesiatov:
The guide itself was merged into a single file for easier navigation.


docs/RenderersGuide.md
```diff
@@ -0,0 +1,366 @@
+---
+authors:
+  - "@carson-katri"
```
carson-katri, line 3:
```suggestion
  - [@carson-katri](https://github.com/carson-katri)
```

MaxDesiatov: It's just a frontmatter key, which isn't displayed in the rendered document. I'll move it out of the frontmatter block to render properly.

carson-katri: Oh ok that's fine then.

PR #252, opened 2020-08-05
Use the latest 5.3 snapshot in `.swift-version`

MaxDesiatov:
These SwiftWasm snapshots should be more stable in general (as they are made off the 5.3 branch, which is slowly converging to a release this fall) and also have a workaround for https://github.com/swiftwasm/JavaScriptKit/issues/6 included. They still use the old metadata layout, so Runtime and OpenCombine dependencies had to be updated in `Package.swift` for `@ObservableObject` to work with these snapshots.


Package.swift
```diff
@@ -38,8 +38,8 @@ let package = Package(
     // Dependencies declare other packages that this package depends on.
     // .package(url: /* package url */, from: "1.0.0"),
     .package(url: "https://github.com/kateinoigakukun/JavaScriptKit.git", .revision("c90e82f")),
-    .package(url: "https://github.com/MaxDesiatov/Runtime.git", .branch("wasi-build")),
-    .package(url: "https://github.com/MaxDesiatov/OpenCombine.git", .branch("observable-object")),
+    .package(url: "https://github.com/MaxDesiatov/Runtime.git", .branch("wasi-build-5.3")),
+    .package(url: "https://github.com/MaxDesiatov/OpenCombine.git", .branch("observable-object-5.3")),
```
ie-ahm-robox, line 42: - :no_entry_sign: Line should be 100 characters or less: currently 102 characters (`line_length`)

PR #253, opened 2020-08-05
Attempt `padding` modifier fusion to avoid nested `div`s

MaxDesiatov:
This allows fusing nested `.padding` modifiers into a single `div` that sums up padding values from all these modifiers.

Before:

```swift
Text("text").padding(10).padding(20)
```

rendered to this (text styling omitted for brevity):

```html
<div style="padding-top: 20.0px; padding-left: 20.0px; padding-bottom: 20.0px; padding-right: 20.0px;">
  <div style="padding-top: 10.0px; padding-left: 10.0px; padding-bottom: 10.0px; padding-right: 10.0px;">
    <span>text</span>
  </div>
</div>
```

Now it renders as

```html
<div style="padding-top: 30.0px; padding-left: 30.0px; padding-bottom: 30.0px; padding-right: 30.0px;">
  <span>text</span>
</div>
```

I hope this approach could be applied to other modifier combinations where it makes sense (in separate PRs).


Sources/TokamakCore/Reflection/Pointers/MetadataOffset.swift
```diff
@@ -25,7 +25,7 @@ struct MetadataOffset<Pointee> {
 
   func apply(to ptr: UnsafeRawPointer) -> UnsafePointer<Pointee> {
     #if arch(wasm32)
-    return unsafeBitCast(offset, to: UnsafePointer<Pointee>.self)
+    return UnsafePointer<Pointee>(bitPattern: Int(offset))!
```
MaxDesiatov, line 28: This fixes a warning that was previously produced on every build.

foscomputerservices (approved):
Very nice to have tests for these now. ✅

PR #259, opened 2020-08-14
Fix `NavigationView` broken state after re-render

MaxDesiatov:
After #241 was merged, `NavigationView` started using `@ObservedObject` instead of `@State` for its destination storage. This uncovered a bug where `@ObservedObject` properties that have their values provided in-place lost their subscriptions after a re-renderer. This manifested in navigation links not working in the demo after any scene phase changes that trigger a re-render. This was caused by a new `NavigationContext` being created on every render, but a corresponding `MountedCompositeElement` instance kept dangling subscription to the old `NavigationContext` instance, while new subscriptions weren't created.

This PR splits subscriptions on `MountedCompositeElement` into separate "transient" and "persistent" arrays. Persistent subscriptions are set up once when an element is mounted, and are automatically cleaned up when the element is unmounted. Scene phase and color scheme subscriptions on `MountedApp` are now primary examples of these persistent subscriptions, but `@StateObject` is another future use case to be implemented in a separate PR. Transient subscriptions are recreated on every re-render to allow new `@ObservedObject` instances to establish their subscriptions.

`NavigationView` in the demo still resets to the empty destination view when the scene phase observer is triggered, but `NavigationLink` selections are fixed now. Resetting to the empty destination view on scene phase changes will be fixed when `@StateObject` implementation is available in a separate PR.

### DIFF
--- a/Sources/TokamakCore/MountedViews/MountedCompositeElement.swift
+++ b/Sources/TokamakCore/MountedViews/MountedCompositeElement.swift
@@ -20,8 +20,22 @@ import CombineShim
 class MountedCompositeElement<R: Renderer>: MountedElement<R> {
   let parentTarget: R.TargetType
 
+  /** An array that stores type-erased values captured with the `@State` property wrappers used
+   in declarations of this element.
+   */
   var state = [Any]()
-  var subscriptions = [AnyCancellable]()
+
+  /** An array that stores subscriptions to updates on `@ObservableObject` property wrappers used
+    in declarations of this element. These subscriptions are transient and may be cleaned up on
+    every re-render of this composite element.
+   */
+  var transientSubscriptions = [AnyCancellable]()
+
+  /** An array that stores subscriptions to updates on `@StateObject` property wrappers and renderer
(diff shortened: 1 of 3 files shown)


MaxDesiatov:
They were persistent initially, but as I mentioned in the description, when re-rendered new `ObservableObject` instances are created, which invalidates old subscriptions.

j-f1:
> when re-rendered new `ObservableObject` instances are created, which invalidates old subscriptions.

Is that intended behavior or a deviation from SwiftUI?

MaxDesiatov:
That's intended and is the reason for adding `@StateObject` this year to the upstream API, which persists across re-renders.

MaxDesiatov:
BTW, if the instance is created outside of the element hierarchy or in the parent element that's not re-rendered, it persists. That was the recommended way to work around it, but `NavigationView` needs to instantiate the context on its own.

Also, more details are available here: https://stackoverflow.com/questions/62544115/what-is-the-difference-between-observedobject-and-stateobject-in-swiftui

PR #260, opened 2020-08-15
Implement `StateObject` property wrapper

### LINKED ISSUE #158: `@StateObject` support
Interesting gotcha described here: https://www.mattmoriarity.com/2020-07-03-stateobject-and-observableobject-in-swiftui/

MaxDesiatov:
Resolves #158.

Fixes a bug where `NavigationView` destination was reset after scene phase changes (or any re-renders caused by environment changes for that matter).

This was caused by `@ObservedObject` destination being recreated, now `@StateObject` persists it across re-renders.

The `setter` property of the `ValueStorage` protocol is now moved to a separate `WritableValueStorage` protocol. The reasoning is that `StateObject` doesn't need its wrapped value to be set directly as it operates on it by reference, not by value, thus `StateObject` doesn't need any wrapped value setters.

### DIFF
--- a/Sources/TokamakCore/MountedViews/MountedCompositeElement.swift
+++ b/Sources/TokamakCore/MountedViews/MountedCompositeElement.swift
@@ -20,10 +20,10 @@ import CombineShim
 class MountedCompositeElement<R: Renderer>: MountedElement<R> {
   let parentTarget: R.TargetType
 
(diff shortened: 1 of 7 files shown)


MaxDesiatov:
Yes please, a separate issue would definitely help in tracking this 👍

PR #262, opened 2020-08-17
Set versions of dependencies in `Package.swift`

MaxDesiatov:
This allows specifying 0.3 (to be released after this PR is merged) dependency on Tokamak in `carton` templates, otherwise branch/commit dependencies in our `Package.swift` can't be correctly resolved.

`swift test` is temporarily disabled on macOS as the upstream Swift toolchain doesn't support unsafe flags in the JavaScriptKit dependency together with a `from` constraint. We could run it on Linux, but my OpenCombine fork doesn't support Linux builds yet (logged as #263).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
         run: |
           set -ex
           sudo xcode-select --switch /Applications/Xcode_12_beta.app/Contents/Developer/
-          swift test
+          # swift test
           xcodebuild -version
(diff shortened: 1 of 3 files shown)


MaxDesiatov:
@j-f1 I hope to tag new releases of both `Tokamak` and `carton` in the next few days, so merging this early to test `carton` templates. LMK if you have any objections, I can revert or submit a follow-up PR.

j-f1:
LGTM! Was on vacation so couldn’t review.

PR #265, opened 2020-08-19
Add missing JavaScriptKit import to `README.md`

MaxDesiatov:
The import is currently missing, which may be confusing, especially if example code is copied without any context.


carson-katri (dismissed):

MaxDesiatov:
A link to JSKit repository was missing too.

PR #271, opened 2020-08-30
Allow non-body mount host node

kateinoigakukun:
I'm creating Tokamak playground app for demonstration and it requires non-body hosting.

Screen Shot 2020-08-30 at 10 48 43

### DIFF
--- a/Sources/TokamakDOM/App/App.swift
+++ b/Sources/TokamakDOM/App/App.swift
@@ -21,6 +21,9 @@ import TokamakCore
 import TokamakStaticHTML
 
 extension App {
+  public static func _launch(_ app: Self, _ rootEnvironment: EnvironmentValues) {
+    _launch(app, rootEnvironment, TokamakDOM.body)
+  }
   /// The default implementation of `launch` for a `TokamakDOM` app.
   ///
   /// Creates a host `div` node and appends it to the body.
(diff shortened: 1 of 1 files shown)


MaxDesiatov (approved):
LGTM 👍

Sources/TokamakDOM/App/App.swift
```diff
@@ -21,15 +21,17 @@ import TokamakCore
...
   /// The default implementation of `launch` for a `TokamakDOM` app.
   ///
   /// Creates a host `div` node and appends it to the body.
   ///
   /// The body is styled with `margin: 0;` to match the `SwiftUI` layout
   /// system as closely as possible
   ///
-  public static func _launch(_ app: Self, _ rootEnvironment: EnvironmentValues) {
-    let body = TokamakDOM.body
+  public static func _launch(_ app: Self, _ rootEnvironment: EnvironmentValues, _ body: JSObjectRef) {
```
j-f1, line 34: This would eliminate the need to have two separate functions.

```suggestion
  public static func _launch(_ app: Self, _ rootEnvironment: EnvironmentValues, _ body: JSObjectRef = TokamakDOM.body) {
```

kateinoigakukun: `TokamakDOM.body` can't be appeared in default value because it's access level is internal.

MaxDesiatov: I wonder if making it `public` would be ok, although `body` is a name too short, which could cause name collisions. There's no problem in resolving them though by spelling out `TokamakDOM.body` fully, WDYT?

kateinoigakukun: IMO, we shouldn't expose extra API to shorten the internal code.

PR #273, opened 2020-09-02
Support `spacing` property on `HStack`/`VStack`

### LINKED ISSUE #272: `spacing` initializer argument has no effect in `VStack`
Verified in Chrome, Safari, and Firefox, elements in a `VStack` always stick to each other regardless of the `spacing` value:

```swift
struct VStackTest: View {
  @State private var email: String = ""
  @State private var password: String = ""

  @State private var isSubmitted = false

  var body: some View {
    VStack(spacing: 100) {
      TextField("Email", text: $email)
      SecureField("Password", text: $password)
      Button("Submit") { isSubmitted = true }
    }
  }
}
```

MaxDesiatov:
It's much easier to implement stack spacing when stacks are rendered as single-row or single-column grids, and [grid gaps already work in all browsers](https://caniuse.com/#feat=mdn-css_properties_gap_grid_context). For this we need to slightly bump browser version requirements, most notably from Safari 11 to Safari 12.

Resolves #272.


j-f1:
Definitely reasonable to use grids for Stacks at least until `gap` is fully supported!

carson-katri:
Yeah, I have no problems with using grid.

souporserious:
CSS Grid is a great way to support child spacing! I wrote about using them in React [here](https://souporserious.com/stack-components-with-css-grid/). Would love to help out with this if you're looking for contributors 🙂

MaxDesiatov:
We're always looking for contributors, thank you for the offer! Please feel free to open a PR even if you just started working on this. You can mark it as a draft even if it's work in progress, that will let everyone know what you're working on. Just open an issue if you see any problems or need any help! 🙂

PR #274, opened 2020-09-03
Add basic `Form` implementation

MaxDesiatov:
Currently blocked by #272. On mobile it obviously doesn't look the way it looks on iOS, but I think on desktop it's close enough to looking like a normal form. I wonder if we should use `<label>` tags instead of placeholders for a11y here? Also, would adding a real `<form>` tag make any difference, especially from the a11y perspective?


MaxDesiatov:
MDN suggests using the `<form>` tag instead: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Form_Role

PR #275, opened 2020-09-14
Re-export `HTML` type in `TokamakDOM`

kateinoigakukun:
`HTML` should be available when targeting dynamic HTML. Without this symlink, we need to import only `HTML` to avoid protocol conformance conflict.

e.g. https://github.com/kateinoigakukun/swiftwasm-pad/blob/9dc42e0ac6c37f6a408fdf5cbd37358b5eebb070/Frontend/Sources/SwiftWasmPad/App.swift#L4

### DIFF
--- a/Sources/TokamakDOM/App/App.swift
+++ b/Sources/TokamakDOM/App/App.swift
@@ -24,6 +24,7 @@ extension App {
   public static func _launch(_ app: Self, _ rootEnvironment: EnvironmentValues) {
     _launch(app, rootEnvironment, TokamakDOM.body)
   }
+
   /// The default implementation of `launch` for a `TokamakDOM` app.
   ///
   /// Creates a host `div` node and appends it to the body.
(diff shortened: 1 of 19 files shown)


MaxDesiatov (commented):
Wouldn't it be more maintainable if `TokamakDOM` re-exported the symbol with `public typealias HTML = TokamakStaticHTML.HTML`, for example in `DynamicHTML.swift` or wherever is more suitable?

kateinoigakukun:
@MaxDesiatov As far as I understand, `TokamakDOM` and `TokamakStaticHTML` should be mutually exclusive. Is this incorrect understanding?

MaxDesiatov:
They aren't mutually exclusive, `TokamakDOM` depends on `TokamakStaticHTML` and extends it to avoid duplication.

carson-katri (approved):

MaxDesiatov (approved):
Awesome, thank you!

MaxDesiatov:
@kateinoigakukun I've sent you an invitation to the Tokamak team, feel free to merge 🙂

MaxDesiatov (approved):

kateinoigakukun:
Please merge this PR 🙏  I don't have commit permission 🤔

MaxDesiatov:
Sorry about that, I've updated the permissions, feel free to merge and let me know if there are any problems.

PR #276, opened 2020-09-17
Add Link view, update JavaScriptKit to 0.8.0

carson-katri:
The `Link` View is added in this PR. I also upgraded JavaScriptKit to `0.8.0`.

> NOTE: Publish integration is being moved to [`TokamakUI/TokamakPublish`](https://github.com/TokamakUI/TokamakPublish)

This adds the ability to use `TokamakStaticHTML` in place of `Plot`.

A `tokamakFoundation` theme is provided, and custom ones can be created with the `TokamakHTMLFactory` protocol.

That requires implementing the following functions:
```swift
func makeIndexView(for index: Index, context: PublishingContext<Self.Site>) throws -> IndexView
func makeSectionView(
    for section: Publish.Section<Self.Site>,
    context: PublishingContext<Self.Site>
  ) throws -> SectionView
func makeItemView(for item: Item<Self.Site>, context: PublishingContext<Self.Site>) throws
    -> ItemView
func makePageView(for page: Page, context: PublishingContext<Self.Site>) throws -> PageView
func makeTagListView(for page: TagListPage, context: PublishingContext<Self.Site>) throws
    -> TagListView
func makeTagDetailsView(for page: TagDetailsPage, context: PublishingContext<Self.Site>) throws
    -> TagDetailsView
```
These will be rendered to HTML, and embedded in `Plot.HTML` elements.

You can then declare your theme with an extension on the `Theme` struct:
```swift
extension Theme {
  public static var yourThemeName: Self {
    Theme(htmlFactory: FoundationTokamakFactory())
  }
}
```

If you haven't used Publish before, you can install the CLI with:
```sh
git clone https://github.com/JohnSundell/Publish.git
cd Publish
make
```
Then run `publish new` and add `TokamakStaticHTML` as a dependency.


Sources/TokamakStaticHTML/PublishSupport/FoundationTokamakFactory+Views.swift
```diff
@@ -0,0 +1,124 @@
...
+                string: context.sections[section].path
+                  .absoluteString
+              )!) {
+                Text(context.sections[section].title)
+                  .underline(section == selectedSection, color: .primary)
+              }
+            }
+          }
+        }
+        // FIXME: Could not cast value of type 'Swift.Optional<TokamakCore.HStack<TokamakCore.ForEach<Swift.Array<PublishTokamak.PublishTokamak.SectionID>, PublishTokamak.PublishTokamak.SectionID, TokamakCore.ModifiedContent<TokamakCore.ModifiedContent<TokamakCore.ModifiedContent<TokamakCore.Link<TokamakCore.Text>, TokamakCore._PaddingLayout>, TokamakCore._BackgroundModifier<TokamakCore.Color>>, TokamakCore._PaddingLayout>>>>' (0x7ffac7021410) to 'TokamakCore.ViewDeferredToRenderer' (0x7ffac7021710).
```
ie-ahm-robox, line 72: - :no_entry_sign: Line should be 100 characters or less: currently 505 characters (`line_length`)

MaxDesiatov:
Fabulous! I was thinking of proposing something like this literally a few days ago, so glad this will be resolved soon 👏

Package.swift
```diff
@@ -4,6 +4,16 @@
 
 import PackageDescription
 
+#if os(WASI)
```
MaxDesiatov, line 7: Does this actually have the desired effect? `Package.swift` is evaluated on the host platform, not on the target platform, so `#if os(WASI)` condition will never be evaluated until we have a self-hosted toolchain running in the browser (we probably won't for long time). These kind of exclusion of dependencies should be implemented with `condition: .when(platforms: [.wasi])`, as done for the `TokamakShim` target below.

This check also seems to be the reason for the Linux build failing.

carson-katri: Oops! 😂 

Although conditions don't seem to work, as we can't depend on a version # of JavaScriptKit since it has unsafe build flags, unless I'm doing something wrong? Should I just change JavaScriptKit to a commit?

MaxDesiatov: Are you using a recent 5.3 SwiftWasm snapshot as recorded in the `.swift-version` file in `main`? These snapshots allow adding JSKit with a version as a dependency, even with unsafe flags.

MaxDesiatov: I see what you mean, Swift toolchains for other platforms don't have the unsafe flags workaround. I hope that if you don't attempt to build products that depend on JavaScriptKit, it won't complain about unsafe flags? As in `swift build --product TokamakStaticHTML`, or when you build something that depends on Tokamak, but doesn't depend on `TokamakDOM`?

carson-katri: In my publish test project it seems to resolve the dependencies for all the products, not just TokamakStaticHTML...

MaxDesiatov: That's probably because https://github.com/apple/swift-package-manager/pull/2749 was merged after Swift 5.3 was branched off. Guess we're stuck with this behavior until the next version of Swift. 😞

IDK what's the best workaround here, maybe we could get rid of unsafe flags in JavaScriptKit somehow?

@kateinoigakukun what do you think? The gist is that packages that depend on JavaScriptKit can't be used on non-Wasm platforms due to the use of unsafe flags. That's because Swift 5.3 resolves all dependencies regardless of whether they're used or not. It even ignores platform conditions apparently. Do you think we could find a way to build JavaScriptKit without unsafe flags by modifying our SwiftWasm 5.3 branch? Maybe we could make it infer linker flags automatically?

kateinoigakukun: I think we need to add new `LinkerSetting` cases in SwiftPM to support exporting and importing symbols.

MaxDesiatov: Even if we add new `LinkerSetting` values, we still need upstream to support them. Now is the best time to submit a PR with these settings upstream before the new major version is branched off, but it means that we still can't create a package manifest that declares these new safe linker settings consumable with upstream Swift 5.3. Or am I missing something?

kateinoigakukun: The only way we can take to avoid this issue is to separate this repository into a package that depends on JavaScriptKit and a package that doesn't depends on JavaScriptKit.

MaxDesiatov: As far as I know, Swift 5.3 still downloads the whole tree of dependencies without checking whether they're used or not. This was fixed in apple/swift-package-manager#2749, but it was merged after Swift 5.3 was branched off.

kateinoigakukun: I can't figure out another idea 🤷 The user must use wasm toolchain now.

MaxDesiatov: A horrible hack would be to patch SwiftPM to check whether it's linking JavaScriptKit and to inject the required linker flags. This probably isn't worse than what we do now, as SwiftPM in 5.3 SwiftWasm snapshots already checks for a package name and allows unsafe flags if it's JavaScriptKit. But submitting a PR upstream to allow more linker flags is a priority then. I don't want to miss the next version cut-off again. 🙏

carson-katri:
@MaxDesiatov this builds and runs with the latest `carton`. Does the `swiftwasm-action` need to be updated to a newer snapshot?

MaxDesiatov:
Yes, I'll do that in the next few days, there are a few more changes I want to land in SwiftPM to allow OpenCombine tests to run. If it's urgent for you, please feel free to submit a PR to the `swiftwasm-action` repo, you probably also have enough permissions to upload a new image yourself after that PR is merged. Or maybe we should automate that if it's not done yet.

carson-katri:
No rush 🙃

MaxDesiatov:
Sorry for the delay, I'm aware that we have too many stalled PRs and I want to get it all moving again. It's just toolchain and `carton` changes consumed too much of my development capacity...

I also hope I'm not completely blocking everything. If there's anything you can't do without me, but you'd like to proceed on your own anyway, please feel free to ping me. I want our projects to be as distributed as possible without bottlenecks in communication and other dev processes.

MaxDesiatov:
Another thing I had in mind for this PR: does this Publish integration require access to anything private or internal in Tokamak? Would it work better as a separate package, and a Publish plugin?

carson-katri:
It can definitely work as a separate package. As for making it a Publish plugin, those aren't really meant for building the theme from my understanding, so I don't think it'd work as one.

MaxDesiatov:
Cool, feel free to create a repo for the package in this org if you'd like. I think a separate repo would be more suitable. It would allow specifying a dependency on a specific version range of Publish.

It would also mean we can add support for other static generators in the future separately. Otherwise we'd need to maintain all other integrations within this repo, or Publish would be the only privileged one.

carson-katri:
Would `TokamakPublish` work as the name?

MaxDesiatov:
LGTM 👍

carson-katri:
Keeping this PR open because it adds the `Link` view and JSKit still needs to be upgraded to 0.8.0 so it can build.

MaxDesiatov:
I've uploaded new `swiftwasm-action` and `carton` Docker images with the latest toolchain just now, in case those are still needed.

carson-katri:
Awesome, CI is now passing 👍

MaxDesiatov (approved):
LGTM 👍

MaxDesiatov:
@carson-katri is this ready to merge?

PR #281, opened 2020-09-28
Fix compatibility with JavaScriptKit 0.7

MaxDesiatov:
This PR requires `carton` 0.6.0 that you can install from Homebrew as usual.

To cleanly manage scheduler closures, new `JSScheduler` class is introduced that conforms to OpenCombine's `Scheduler` protocol. I think it will be moved to OpenCombineJS in the future.

~Basic things seem to work, but I think there's a noticeable performance regression somewhere. When I switch between navigation views, it takes almost a second to load everything. I've been staring at profiling data for a bit, `JSString` which was introduced in JavaScriptKit 0.7 is there somewhere, but I'm not sure what exactly is the culprit yet.~
Update: I no longer can get the performance regression reproduced. 🤷‍♂️

### DIFF
--- a/.swift-version
+++ b/.swift-version
@@ -1,1 +1,1 @@
-wasm-5.3-SNAPSHOT-2020-07-27-a
+wasm-5.3-SNAPSHOT-2020-09-25-a
--- a/Package.resolved
+++ b/Package.resolved
@@ -3,11 +3,11 @@
     "pins": [
       {
         "package": "JavaScriptKit",
-        "repositoryURL": "https://github.com/kateinoigakukun/JavaScriptKit.git",
+        "repositoryURL": "https://github.com/swiftwasm/JavaScriptKit.git",
         "state": {
           "branch": null,
-          "revision": "c90e82fe1d576a2ccd1aae798380bf80be7885fb",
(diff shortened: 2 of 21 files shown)


MaxDesiatov:
Here are the profiling results, you shouldn't compare absolute numbers, but relative time consumption among functions.

In the `main` branch protocol conformance consumes about half of all processing time:
Screenshot 2020-09-28 at 19 24 15

In the branch for this PR `swjs_call_function_with_this` becomes the top consumer:

Screenshot 2020-09-28 at 19 53 43

I'm still not 100% sure it comes from the JavaScriptKit update, but I don't see any scheduler related calls consuming any significant amount of time, and the scheduler and JSKit version bump are the only major things that changed here.

MaxDesiatov:
I can no longer reproduce the performance regression. `carton` 0.6.0 is now avialable on Homebrew, so I'd appreciate if you try things on your side. Regardless, it's time for setting up performance testing on CI...

MaxDesiatov:
The CI failure was caused by some issues with GitHub Actions, now ready for review 🙂

kateinoigakukun (approved):

PR #288, opened 2020-10-06
Add `URLHashDemo` w/ `window.onhashchange` closure

### LINKED ISSUE #284: Pattern for handling global dom events
I'm looking for a pattern for handling global javascript/dom events, such as handling back button navigation ("window.onpopstate"). 
The reason was to workaround navigation problems with wasm apps, with `pushState` and `popState`. 

I tried various scenarious, based on `https://github.com/TokamakUI/Tokamak/blob/main/Sources/TokamakDOM/App/ColorSchemeObserver.swift`, and: 

-  it compiles fine
- the event defined in `JSClosure` is fired, but when I'm sending new value to publisher, it's not received by destination `View` no matter if I use `ObservedObject`, or low level publisher + sink. 

Looks like some magic sauce is needed to communicate from JSClosure which I'm missing.  

Can you provide some example how to do it?

MaxDesiatov:
Resolves #284.

### DIFF
--- a/Sources/TokamakDemo/TokamakDemo.swift
+++ b/Sources/TokamakDemo/TokamakDemo.swift
@@ -144,6 +144,7 @@ struct TokamakDemoView: View {
           #if os(WASI)
           Section(header: Text("TokamakDOM")) {
             NavItem("DOM reference", destination: DOMRefDemo())
+            NavItem("URL hash changes", destination: URLHashDemo())
           }
           #endif
         }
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@vi4m please check out the code in this PR. I imagine `onpopstate` would work similarly. The key bit here is that we need to manage lifetime of closures manually, because JavaScript doesn't have finalizers. Swift code can't be notified when JavaScript references to these closures are deallocated, so everything need to be cleaned up manually. This is possible thanks to `HashState` being a reference type with `deinit`.

vi4m:
Thanks! This code looks really easy on the eyes, and works perfectly. 
I'm sure it will enrich documentation / help other early adopters as well 👍

carson-katri (approved):

PR #289, opened 2020-10-06
Add @carson-katri and @kateinoigakukun to `FUNDING.yml`

MaxDesiatov:
The sponsors list is sorted alphabetically 🙂

@j-f1 if/when you have a sponsorship account set up, or would like to take payments through some other means, plase feel free to create a PR or suggest anything here 👍


carson-katri (approved):
Awesome, thank you!

j-f1 (approved):
LGTM! Not looking for funding at the moment but will definitely update if/when that changes.

kateinoigakukun (approved):

PR #297, opened 2020-11-08
Xcode compatibility

kateinoigakukun:
Now, we can develop Tokamak app project with Xcode.


MaxDesiatov (approved):
Thank you for sorting this out!

Issue #294, opened 2020-10-29
Shuffled order of components


vi4m:
I've found some interesting behavior related to Views drawing order. 

I'd like to show "Add", "List" views depending on the window location hash. 
In this example code components order are not deterministic. Is this because of some Pub/Sub delay? 
![image](https://user-images.githubusercontent.com/552398/97591456-edafac00-19ff-11eb-908c-7148dc84c897.png)
1) When i visit it the first time, it works 100% correctly

![Zrzut ekranu 2020-10-29 o 15 58 02](https://user-images.githubusercontent.com/552398/97591744-35cece80-1a00-11eb-96d2-b4ddf12db865.png)

2) After clicking "list", it should show "begin" + list + "end", but the order is messed up.
![Zrzut ekranu 2020-10-29 o 15 58 07](https://user-images.githubusercontent.com/552398/97591934-69115d80-1a00-11eb-8341-264944489ba4.png)

I managed to put reproducible case in one file here:

https://gist.github.com/vi4m/300dffe461811aa96dae210929ddb5e5

MaxDesiatov:
@vi4m thanks for the detailed report! I'll be able to have a look in the next few days, unless someone else picks this up in the meantime.

MaxDesiatov:
Thanks again for raising this, and sorry for the delay. I've narrowed it down to this very simple case:

```swift
struct Choice: View {
  @State private var choice = false

  var body: some View {
    HStack {
      Button("Trigger") {
        choice = true
      }
      if choice {
        Text("true")
      } else {
        VStack {
          Text("false")
        }
      }
      Text("end")
    }
  }
}
```

The fact that the `Text` element in the mounted views tree is replaced with a `VStack` currently causes these elements to be remounted. The former is removed and the latter is added with [`insertAdjacentHTML`](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML). But during the insertion of the new element it doesn't have access to its siblings, only to its parent, and we mount it with the `beforeend` argument passed to `insertAdjacentHTML`. It inserts the element as the last child of its parent. That is clearly wrong. I currently see two possible solutions for it:

1. Allow elements to reference its previous siblings during remounting and use `insertAdjacentHTML` with `afterend` on the sibling instead of `beforeend` on the parent. I've started with this option and it touches too much of our renderer and reconciler code and complicates things significantly.
[...]

PR #301, opened 2020-11-10
Pass sibling to `Renderer.mount`, fix update order

MaxDesiatov:
Resolves, but adds no tests cases to the test suite for #294. See the issue for the detailed description of the problem.

I will add end-to-end tests for this in future PRs.

I've tested these cases manually so far:

```swift
struct Choice: View {
  @State private var choice = false

  var body: some View {
    HStack {
      Button("Trigger") {
        choice.toggle()
      }
      if choice {
        Group {
          Text("true")
          Text("true")
        }
      } else {
        VStack {
          Text("false")
        }
      }
      Text("end")
    }
  }
}
```

Note the `Group` view with multiple children in this one, it uncovered required checks for `GroupView` conformance.

Also tested these more simple cases:

```swift
struct Choice: View {
  @State private var choice = false

  var body: some View {
    HStack {
      Button("Trigger") {
        choice.toggle()
      }
      if choice {
        Group {
          // single child
          Text("true")
        }
      } else {
        VStack {
          Text("false")
        }
      }
      Text("end")
    }
  }
}
```

and

```swift
struct Choice: View {
  @State private var choice = false

  var body: some View {
    HStack {
      Button("Trigger") {
        choice.toggle()
      }
      if choice {
        // single child, no nesting
        Text("true")
      } else {
        VStack {
          Text("false")
        }
      }
      Text("end")
    }
  }
}
```


Sources/TokamakCore/MountedViews/MountedApp.swift
```diff
@@ -22,12 +22,12 @@ import Runtime
 // is the computed content of the specified `Scene`, instead of having child
 // `View`s
 final class MountedApp<R: Renderer>: MountedCompositeElement<R> {
-  override func mount(with reconciler: StackReconciler<R>) {
+  override func mount(before _: R.TargetType? = nil, with reconciler: StackReconciler<R>) {
```
MaxDesiatov, line 25: `App` elements have no siblings

Sources/TokamakCore/Renderer.swift
```diff
@@ -31,9 +31,6 @@ public protocol Renderer: AnyObject {
    */
   associatedtype TargetType: Target
 
-  /// Reconciler instance used by this renderer.
-  var reconciler: StackReconciler<Self>? { get }
```
MaxDesiatov, line 35: It wasn't referenced anywhere, so no need to have it as a requirement. It also allowed making this property completely `private` in `Renderer` implementations.

Sources/TokamakDOM/DOMRenderer.swift
```diff
@@ -71,7 +71,7 @@ func appendRootStyle(_ rootNode: JSObject) {
 }
 
 final class DOMRenderer: Renderer {
-  private(set) var reconciler: StackReconciler<DOMRenderer>?
+  private var reconciler: StackReconciler<DOMRenderer>?
```
MaxDesiatov, line 74: Still keeping the property implementation itself around, as we need a non-zero retain count on it. The reconciler is really what drives the rendering process.

carson-katri (approved):
Looks good! The original issue was a great catch 😅

PR #307, opened 2020-11-27
Add Preferences

carson-katri:
This adds the `PreferenceKey` protocol and related modifiers.

There are a few bugs still, but I wanted to get some feedback.

### DIFF
--- a/Sources/TokamakCore/MountedViews/MountedElement.swift
+++ b/Sources/TokamakCore/MountedViews/MountedElement.swift
@@ -76,10 +76,10 @@ public class MountedElement<R: Renderer> {
   var typeConstructorName: String {
     switch element {
     case .app: fatalError("""
-      `App` values aren't supposed to be reconciled, thus the type constructor name is not stored \
-      for `App` elements. Please report this crash as a bug at \
-      https://github.com/swiftwasm/Tokamak/issues/new
-      """)
+    `App` values aren't supposed to be reconciled, thus the type constructor name is not stored \
+    for `App` elements. Please report this crash as a bug at \
+    https://github.com/swiftwasm/Tokamak/issues/new
+    """)
     case let .scene(scene): return scene.typeConstructorName
     case let .view(view): return view.typeConstructorName
     }
@@ -87,21 +87,34 @@ public class MountedElement<R: Renderer> {
 
   var mountedChildren = [MountedElement<R>]()
   var environmentValues: EnvironmentValues
+  unowned var parent: MountedElement<R>?
+  /// `didSet` on this field propagates the preference changes up the view tree.
+  var preferenceStore: _PreferenceStore {
+    didSet {
+      parent?.preferenceStore.merge(with: preferenceStore)
+    }
+  }
 
-  init(_ app: _AnyApp, _ environmentValues: EnvironmentValues) {
+  init(_ app: _AnyApp, _ environmentValues: EnvironmentValues, _ parent: MountedElement<R>?) {
     element = .app(app)
+    preferenceStore = .init()
+    self.parent = parent
     self.environmentValues = environmentValues
     updateEnvironment()
   }
 
-  init(_ scene: _AnyScene, _ environmentValues: EnvironmentValues) {
+  init(_ scene: _AnyScene, _ environmentValues: EnvironmentValues, _ parent: MountedElement<R>?) {
     element = .scene(scene)
+    preferenceStore = .init()
+    self.parent = parent
     self.environmentValues = environmentValues
     updateEnvironment()
   }
 
-  init(_ view: AnyView, _ environmentValues: EnvironmentValues) {
+  init(_ view: AnyView, _ environmentValues: EnvironmentValues, _ parent: MountedElement<R>?) {
     element = .view(view)
+    preferenceStore = .init()
+    self.parent = parent
     self.environmentValues = environmentValues
     updateEnvironment()
   }
@@ -122,7 +135,11 @@ public class MountedElement<R: Renderer> {
     return info
   }
 
-  func mount(before sibling: R.TargetType? = nil, with reconciler: StackReconciler<R>) {
+  func mount(
+    before sibling: R.TargetType? = nil,
+    on parent: MountedElement<R>? = nil,
+    with reconciler: StackReconciler<R>
+  ) {
     fatalError("implement \(#function) in subclass")
   }
 
@@ -157,7 +174,7 @@ extension TypeInfo {
     // swiftlint:disable force_try
     // Extract the view from the AnyView for modification, apply Environment changes:
     if genericTypes.contains(where: { $0 is EnvironmentModifier.Type }),
-       let modifier = try! property(named: "modifier").get(from: element) as? EnvironmentModifier
+      let modifier = try! property(named: "modifier").get(from: element) as? EnvironmentModifier
     {
       modifier.modifyEnvironment(&modifiedEnv)
     }
@@ -217,14 +234,15 @@ extension TypeInfo {
 extension AnyView {
   func makeMountedView<R: Renderer>(
     _ parentTarget: R.TargetType,
-    _ environmentValues: EnvironmentValues
+    _ environmentValues: EnvironmentValues,
+    _ parent: MountedElement<R>?
   ) -> MountedElement<R> {
     if type == EmptyView.self {
-      return MountedEmptyView(self, environmentValues)
+      return MountedEmptyView(self, environmentValues, parent)
     } else if bodyType == Never.self && !(type is ViewDeferredToRenderer.Type) {
-      return MountedHostView(self, parentTarget, environmentValues)
+      return MountedHostView(self, parentTarget, environmentValues, parent)
     } else {
-      return MountedCompositeView(self, parentTarget, environmentValues)
+      return MountedCompositeView(self, parentTarget, environmentValues, parent)
     }
   }
 }
--- a/Sources/TokamakCore/MountedViews/MountedElement.swift
+++ b/Sources/TokamakCore/MountedViews/MountedElement.swift
@@ -76,10 +76,10 @@ public class MountedElement<R: Renderer> {
   var typeConstructorName: String {
     switch element {
     case .app: fatalError("""
-    `App` values aren't supposed to be reconciled, thus the type constructor name is not stored \
-    for `App` elements. Please report this crash as a bug at \
-    https://github.com/swiftwasm/Tokamak/issues/new
-    """)
+      `App` values aren't supposed to be reconciled, thus the type constructor name is not stored \
+      for `App` elements. Please report this crash as a bug at \
+      https://github.com/swiftwasm/Tokamak/issues/new
+      """)
     case let .scene(scene): return scene.typeConstructorName
     case let .view(view): return view.typeConstructorName
     }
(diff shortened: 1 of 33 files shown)


Sources/TokamakCore/MountedViews/MountedElement.swift
```diff
@@ -76,32 +76,45 @@ public class MountedElement<R: Renderer> {
...
+    https://github.com/swiftwasm/Tokamak/issues/new
+    """)
     case let .scene(scene): return scene.typeConstructorName
     case let .view(view): return view.typeConstructorName
     }
   }
 
   var mountedChildren = [MountedElement<R>]()
   var environmentValues: EnvironmentValues
+  unowned var parent: MountedElement<R>?
```
MaxDesiatov, line 90: I never remember what are the reasons to use `unowned` instead of `weak`, I guess the latter is always safer?
```suggestion
  weak var parent: MountedElement<R>?
```

Sources/TokamakCore/Preferences/_PreferenceReadingView.swift
```diff
@@ -0,0 +1,92 @@
...
+  public func _force<T>(
+    _ transform: @escaping (Key.Value) -> T
+  ) -> _PreferenceReadingView<Key, T>
+    where T: View
+  {
+    _PreferenceReadingView(value: self, transform: transform)
+  }
+}
+
+/// Delays the retrieval of a `PreferenceKey.Value` by passing the `_PreferenceValue` to a build function.
```
MaxDesiatov, line 50:
```suggestion
/// Delays the retrieval of a `PreferenceKey.Value` by passing the `_PreferenceValue` to a build
/// function.
```

Sources/TokamakCore/MountedViews/MountedElement.swift
carson-katri: Unowned is for when the thing it references has a lifetime >= it’s own IIRC.

Sources/TokamakCore/MountedViews/MountedElement.swift
```diff
@@ -217,14 +234,15 @@ extension TypeInfo {
 extension AnyView {
   func makeMountedView<R: Renderer>(
     _ parentTarget: R.TargetType,
-    _ environmentValues: EnvironmentValues
+    _ environmentValues: EnvironmentValues,
+    _ parent: MountedElement<R>?
```
MaxDesiatov, line 238: Do we need to be explicit with this, or would `nil` as a default argument make sense here?
```suggestion
    _ parent: MountedElement<R>? = nil
```

Sources/TokamakDOM/DOMRenderer.swift
```diff
@@ -157,7 +157,7 @@ final class DOMRenderer: Renderer {
   ) {
     defer { completion() }
 
-    guard let html = mapAnyView(host.view, transform: { (html: AnyHTML) in html })
+    guard let _ = mapAnyView(host.view, transform: { (html: AnyHTML) in html })
```
MaxDesiatov, line 160:
```suggestion
    guard mapAnyView(host.view, transform: { (html: AnyHTML) in html }) != nil
```

Sources/TokamakDemo/PreferenceKeyDemo.swift
```diff
@@ -0,0 +1,79 @@
...
+      Text("One level up sets the color to green, and so on.")
+      SetColor(3, .purple) {
+        SetColor(2, .green) {
+          SetColor(1, .blue)
+        }
+      }
+    }
+    .padding()
+    .background(testKeyValue)
+//    .preferredColorScheme(colorScheme == .light ? .dark : .light)
```
MaxDesiatov, line 39: *nitpick*: Is this comment needed?

MaxDesiatov:
Maybe the macOS failure is fixed in #308?

Sources/TokamakCore/MountedViews/MountedElement.swift
carson-katri: I guess the argument for no default argument would be that only the root element (when creating the `StackReconciler`) sets it to `nil`. So having it be explicit would catch accidental missed arguments.

MaxDesiatov:
Well, that was an improvement, at least the error message is clear, the new preferences demo file needs to be added to the Xcode project 🙂

MaxDesiatov:
Am I missing anything? This is how it looks for me in all browsers? Do I correctly expect different levels to have different colors?

Screenshot 2020-11-27 at 15 30 11

Sources/TokamakDemo/PreferenceKeyDemo.swift
```diff
@@ -0,0 +1,88 @@
...
+
+    init(_ level: Int, _ color: Color, @ViewBuilder _ content: () -> Content) {
+      self.level = level
+      self.color = color
+      self.content = content()
+    }
+
+    var body: some View {
+      VStack {
+        Text("Level \(level)")
```
MaxDesiatov, line 69: I don't remember if `Color` is `CustomStringConvertible`, but maybe there's some way to specify color name in the text here together with the level?

MaxDesiatov:
It changes colors to this after switching between tabs 🤔 

Screenshot 2020-11-27 at 15 32 52

carson-katri:
Yes, looking into that. The state update after the onPreferenceChange doesn’t seem to trigger a rerender.

And yeah, I implemented CustomDebugStringConvertible on Color.

carson-katri:
When I isolate the demo it works perfectly, but in the full demo the state update isn't triggering a rerender or something... 😕

MaxDesiatov:
Maybe something in the environment or other state is messing it up when preferences are wrapped with other views? Maybe it's something in `NavigationView`? I'd try different ways to isolate it with and without `NavigationView` and check if that yields different results.

carson-katri:
So I tried a couple different things, and found it breaks when its the destination of a `NavigationLink`.

carson-katri:
This is definitely an issue with `NavigationView`/`Link`, because updating state from an `onAppear` in a destination does not update either.

carson-katri:
Preferences and `onAppear` now run after the current render has completed which fixes the state update issue. IIRC SwiftUI throws a runtime warning when state is updated during a render, maybe we should do the same?

MaxDesiatov:
Hm, `State` has these stored properties

```swift
var getter: (() -> Any)?
var setter: ((Any) -> ())?
```

which are set before rendering. Maybe we should keep `get` non-nil before rendering, but make `set` non-nil only after rendering? That way attempts to set `State` during rendering will be equivalent to calling `nil` `setter`, which could then trigger a warning or error?

carson-katri:
I can give it a go.

Sources/TokamakDemo/main.swift
```diff
@@ -26,6 +26,20 @@ struct CustomScene: Scene {
   }
 }
 
+struct CountDemo: View {
```
MaxDesiatov, line 29: Is this view still needed? I don't think it's currently used anywhere

MaxDesiatov:
BTW, no need to add warnings for `State` setter in this same PR, we could get that in separately. 🙂

Sources/TokamakDemo/main.swift
carson-katri: Used that to test the `onAppear` state update, I'll get it removed.

carson-katri:
Ok 👍

SwiftUI shows a runtime issue in Xcode, which I don't think is possible for us to do. I'm not sure how we'd get the `#file` and `#line` to the wrappedValue setter for logging though... I guess that can all be discussed separately.

Sources/TokamakCore/Tokens/Color.swift
```diff
@@ -94,10 +103,10 @@ public class _ConcreteColorBox: AnyColorBox {
 public class _EnvironmentDependentColorBox: AnyColorBox {
   public let resolver: (EnvironmentValues) -> Color
 
-  public static func == (lhs: _EnvironmentDependentColorBox,
-                         rhs: _EnvironmentDependentColorBox) -> Bool
-  {
-    lhs.resolver(EnvironmentValues()) == rhs.resolver(EnvironmentValues())
+  override public func equals(_ other: AnyColorBox) -> Bool {
```
MaxDesiatov, line 106: What's the reason for using `equals` instead of `==` and `Equatable`? It should be added in a corresponding code comment.

Sources/TokamakCore/Tokens/Color.swift
```diff
@@ -61,7 +61,14 @@ public class AnyColorBox: AnyTokenBox {
...
   }
 
-  public static func == (lhs: AnyColorBox, rhs: AnyColorBox) -> Bool { false }
+  public static func == (lhs: AnyColorBox, rhs: AnyColorBox) -> Bool {
+    lhs.equals(rhs)
+  }
+
+  public func equals(_ other: AnyColorBox) -> Bool {
+    fatalError("implement \(#function) in subclass")
+  }
```
MaxDesiatov, line 70: Do these two need to be separate for overriding purposes? Please add a doc comment that clarifies it.

Sources/TokamakDemo/PreferenceKeyDemo.swift
```diff
@@ -0,0 +1,114 @@
...
+  @Environment(\.colorScheme) var colorScheme
+
+  var body: some View {
+//    Text("The key color is: \(testKeyValue.description)")
+//        .foregroundColor(testKeyValue)
+//        .backgroundPreferenceValue(TestPreferenceKey.self) { value in
+//            Text("The key passed to the background is \(value.description)")
+//                .foregroundColor(value)
+//        }
+//        .preference(key: TestPreferenceKey.self, value: .green)
```
MaxDesiatov, line 39: **nitpick:** can this be cleaned up?

MaxDesiatov:
LMK if I'm reviewing too early and should just wait for you to request a review. Or if you prefer "draft" PR state for that?

carson-katri:
I'm fine with the reviews now. The only modifier left is `transformPreference`, so it's fairly close.

carson-katri:
Ok, everything should be implemented 😄 

Here it is side-by-side with macOS:

Screen Shot 2020-12-02 at 9 21 30 PM

carson-katri:
One difference from SwiftUI is a `_DelayedPreferenceView` will run once with the key's default value before the child propagates its preference changes. Not sure how to fix this or if it's even important as it should be updated immediately on the next frame.

MaxDesiatov (approved):
Amazing work, much appreciated! 👏

Sources/TokamakCore/Tokens/Color.swift
```diff
@@ -252,6 +268,16 @@ public extension Color {
   }
 }
 
+extension Color: CustomStringConvertible {
+  public var description: String {
+    if let providerDescription = provider as? CustomStringConvertible {
+      return providerDescription.description
+    } else {
+      return String(describing: self)
```
j-f1, line 276: Just noticed that this will recurse if `provider` doesn’t conform to `CustomStringConvertible`. Should this be `String(describing: provider)`, or should we require all providers conform to `CustomStringConvertible`?

MaxDesiatov: Thanks for pointing it out! @carson-katri WDYT?

carson-katri: Good catch! I think `String(describing: provider)` just in case, but also implement `CustomStringConvertible` on all the providers since they really are only made in TokamakCore, if that makes sense?

PR #327, opened 2020-12-06
Run most of the tests with `carton test`

MaxDesiatov:
`ReconcilerTests.swift` was disabled with `!os(WASI)` due to the lack of expectations support in XCTest for SwiftWasm.


MaxDesiatov:
Keeping this as a draft until https://github.com/swiftwasm/carton/issues/200 is resolved.

PR #331, opened 2020-12-09
Added some missing TokamakDOM/Core type typealiases

foscomputerservices:
Added some missing TokamakDOM/Core type typealiases.  It wasn't clear if there was meant to be parity with TokamakStaticHTML/Core as well or not as there were significant differences.  If these should be added there as well, I'd be happy to do so.

### DIFF
--- a/Sources/TokamakDOM/Core.swift
+++ b/Sources/TokamakDOM/Core.swift
@@ -20,7 +20,9 @@ import TokamakCore
 // MARK: Environment & State
 
 public typealias Environment = TokamakCore.Environment
+public typealias EnvironmentKey = TokamakCore.EnvironmentKey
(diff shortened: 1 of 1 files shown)


j-f1 (approved):
It looks like this is indeed an oversight! All the APIs should be imported by consumers via `TokamakDOM`/`TokamakStaticHTML` — `TokamakCore` is kind of a private API.

j-f1:
For `TokamakStaticHTML`, I believe the general idea is to export all the APIs except for those that are intended to be dynamic (buttons, forms, etc).

MaxDesiatov (approved):
Awesome, thanks! TokamakStaticHTML does lack stateful views, but that's temporary. I'm convinced we should allow them for pre-rendering in the future, just need to find an appropriate way to do that.

PR #333, opened 2020-12-13
Add basic GTK renderer code

MaxDesiatov:
Based on the work discussed in #306.


mortenbekditlevsen:
Should I close #306 and move the discussion in here?

@carson-katri I have a question regarding `WidgetView`:
Currently the `update` implementation destroys the current widget. A side effect of this is that `ModifiedContent` elements (which use `WidgetView`) are destroyed upon update.
I tried implementing a background color modifier, and adding this to a view causes it to be destroyed.

In my own attempt to fix the issue, I have extended `WidgetView` to take both a `build` and an `update` closure - and then from within `ModifiedContent` I implement it be calling `update` on the content.

Here's what the `WidgetView` change looks like:
```
struct WidgetView<Content: View>: View, AnyWidget, ParentView {
  let build: (UnsafeMutablePointer<GtkApplication>) -> UnsafeMutablePointer<GtkWidget>
  let update: ((Widget) -> Void)?
  let content: Content
  let expand: Bool

  init(build: @escaping (UnsafeMutablePointer<GtkApplication>) -> UnsafeMutablePointer<GtkWidget>,
       update: ((Widget) -> Void)? = nil,
       expand: Bool = false,
       @ViewBuilder content: () -> Content)
  {
    self.build = build
    self.expand = expand
    self.content = content()
    self.update = update
  }

  func new(_ application: UnsafeMutablePointer<GtkApplication>) -> UnsafeMutablePointer<GtkWidget> {
    build(application)
  }

  func update(widget: Widget) {
    if case let .widget(w) = widget.storage {
      if let u = self.update {
        u(widget)
      } else {
        // Rebuild from scratch
        widget.destroy()
      }
    }
  }
...
```

And the usage from `ModifiedContent` - I refactored it a bit to be able to handle the two different ways to get an `AnyWidget` in the same way:

```
extension ModifiedContent: ViewDeferredToRenderer where Content: View {
  public var deferredBody: AnyView {
    guard let widgetModifier = modifier as? WidgetModifier else {
        return AnyView(content)
    }
    let anyWidget: AnyWidget
    if let anyView = content as? ViewDeferredToRenderer,
       let _anyWidget = mapAnyView(
        anyView.deferredBody,
        transform: { (widget: AnyWidget) in widget }
       )
    {
        anyWidget = _anyWidget
    } else if let _anyWidget = content as? AnyWidget {
        anyWidget = _anyWidget
    } else {
        return AnyView(content)
    }
    return AnyView(WidgetView {
        let contentWidget = anyWidget.new($0)
        widgetModifier.modify(widget: contentWidget)
        return contentWidget
    }
    update: { widget in
        anyWidget.update(widget: widget)

        // Is it correct to run the modifier again after updating?
        // I assume so since the modifier parameters may have changed.
        if case .widget(let w) = widget.storage {
            widgetModifier.modify(widget: w)
        }
    }
    content: {
        if let parentView = anyWidget as? ParentView, parentView.children.count > 1 {
            ForEach(Array(parentView.children.enumerated()), id: \.offset) { _, view in
                view
            }
        } else if let parentView = anyWidget as? ParentView, parentView.children.count == 1 {
            // Is it a worthwile shortcut to not wrap the children in a ForEach when there is only one?
            parentView.children[0]
        }
    })
  }
}
```

I am not 100% certain that I understand the purpose of `WidgetView`, so I don't know if my fix is sound. I also don't know if the update should be optional (as I made it now, with the fallback to destroying the widget upon update), or if I should require it always and then always forward the update call instead of destroying.

carson-katri:
Yeah, so `WidgetView` is kind of like the `HTML` view in `TokamakDOM`. In some cases conforming to `AnyWidget` just doesn't work well, for instance, when building a widget with children.

I think the `update` closure is a good additon, and I don't have a strong opinion on whether it should be `Optional` or not, but if not then the default behavior should be documented.

mortenbekditlevsen:
I have experimented a bit more, and it appears to make sense to always have it (and never destroy the widget upon update), but have an empty closure of the default, since the only place I've had to do extra work is to forward updates with regards to modifiers.

Should I make PRs against this branch, or would it be better to wait until this PR lands?

Btw: The funniest thing happened yesterday: I searched for some gtk APIs and found the `gtk_measure_widget` API which is part of GTK 4. That got me curious and I looked at the version history. When I saw that GTK 4 has been in development for many years I gave up hope for using that API any time soon. 
Then yesterday evening I saw that GTK 4 was just released! 
What a weird coincidence.

---
Just random thoughts - please ignore if you are busy:
---

This brings up a question about the goal for Tokamak of which I am a bit uncertain:
I know that it's a goal to provide a SwiftUI compatible API to create apps using other UI 'backends'. 
What I don't know is how far it should/could go with regards to implementing the same layout behaviors of SwiftUI.

Again with a reference to objc.io, they seem to gather that SwiftUI has two phases of rendering - first the sizing and then the rendering. I guess that the sizing might not make a whole lot of sense in the DOM world (until the DOM nodes are actually rendered), so there it may not make sense to emulate this behavior.
But for some 'backends' (perhaps including GTK 4 although I am only guessing), it may make sense to take it a step further and see if it's possible to get closer to the layout behavior of SwiftUI too.

So my question - although I know that it's probably kind of vague:
Is it a goal to try and mimick the layout behaviour of SwiftUI?
Might it be possible to extend the API to allow for implementing the sizing behaviour for some 'backends' while skipping it for others?

MaxDesiatov:
> Is it a goal to try and mimick the layout behaviour of SwiftUI?

Yes, in my opinion it is. I think layout algorithms are crucial enough for people to expect that they behave the same as SwftUI does on all platforms that Tokamak supports. I think we can get pretty close or maybe even get full compatibility with the DOM renderer without re-implementing our own layout engine. And that's great, because binary code size is a big concern on the web, I'd rather rely on CSS or maybe a few bits of JS here and there than ship our own layout engine in every web app built with Tokamak. But shipping a built-in layout engine on other platforms is probably more acceptable.

This will probably require modifying the renderer API, and I don't know how yet. But all suggestions are welcome. I'll tag a new version of `carton` in the next few days to get CI passing in this PR. Feel free to target the branch of this PR in your own PRs, they'll change their target branch to `main` automatically after this one is merged.

carson-katri:
Here's a bit of discussion on Gtk4:

https://twitter.com/migueldeicaza/status/1336422305224937474?s=21

MaxDesiatov:
@TokamakUI/core-team anyone available for review?

MaxDesiatov:
Most of the commits here come from @carson-katri, which I approve. If any of the maintainers think that either of these commits or my own commits added on top need to be reverted, let me know. I'm merging this now to unblock the rest of the PRs and to avoid merge conflicts.

PR #339, opened 2020-12-18
Gtk renderer image

mortenbekditlevsen:
Added basic `Image` support for GTK and added an example to the demo app.


Sources/TokamakGTKDemo/main.swift
```diff
@@ -46,13 +46,16 @@ struct PickerDemo: View {
...
+  @State private var count: Int = 0
+
   var body: some Scene {
     WindowGroup("Test Scene") {
       List {
+        Image("logo-header.png", bundle: Bundle.module, label: Text("Tokamak Demo"))
         Counter()
         PickerDemo()
         ForEach(1..<100) {
-          Text("Item #\($0)")
```
mortenbekditlevsen, line 55: Oops, this was a test. I'll remove this line later. :-)

MaxDesiatov:
Apologies, GitHub closed it automatically for some reason, while usually it should've only changed the target branch to `main` after it was merged. I'll re-create this in a separate PR.

PR #340, opened 2020-12-18
Add background modifier, support custom widget updates in `WidgetView`

mortenbekditlevsen:

PR #344, opened 2020-12-26
Add `background` modifier, support widget updates in `WidgetView`

MaxDesiatov:
Re-created https://github.com/TokamakUI/Tokamak/pull/340.


carson-katri (approved):
I wonder why it closed those PRs? Is there a setting on the repo that needs to be changed?

MaxDesiatov:
I didn't see anything like that recently. Maybe it's some kind of cross-repo thing, or caused by the fact that PRs were from different users? 🤔

PR #347, opened 2021-01-02
Build the GTK renderer on Ubuntu on CI

MaxDesiatov:
We should make sure that our reconciler and renderer code also builds on Ubuntu.

This PR also contains some formatting changes, which were done by the latest version of SwiftFormat.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,3 +57,29 @@ jobs:
           brew install gtk+3
 
           make build
+
+  gtk_ubuntu_18_04_build:
+    runs-on: ubuntu-18.04
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build the GTK renderer on Ubuntu 18.04
+        shell: bash
+        run: |
+          set -ex
+          sudo apt-get install libgtk+-3.0 gtk+-3.0
+
+          make build
+
+  gtk_ubuntu_20_04_build:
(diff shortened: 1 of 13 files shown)


mortenbekditlevsen:
Hi @MaxDesiatov ,
I recently tried building on Ubuntu 20.04 and get the same 'CRuntime' import errors.
Do you know what they are?

MaxDesiatov:
I didn't do a very thorough investigation, seems that the Runtime dependency in OpenCombine is causing issues. I'm going to replace it with a plain `Mirror` implementation as an attempt to fix it. Just need to cherry-pick https://github.com/OpenCombine/OpenCombine/pull/201 into our fork of OpenCombine, which I'm going to do later today.

mortenbekditlevsen:
I can now build and run TokamakGTKDemo - and it's great to see it running on Ubuntu!

Currently nothing re-renders when state changes, but I don't know if that has anything to do with the changes to OpenCombine with regards to Runtime.

mortenbekditlevsen (approved):
All workflow and Makefile changes looks good to me.
As I mentioned the TokamakGTKDemo does not re-render at state changes, but I don't know if that has anything to do with the 0.12.0-alpha3 version of OpenCombine or not (since I haven't been able to build on Ubuntu before).

MaxDesiatov:
Makes sense. I think lack of updates needs to be investigated separately.

Issue #322, opened 2020-12-05
Crash when stroking a circle


j-f1:
The following code:

```swift
Circle().stroke(Color.gray)  
```

produces this error. Explicitly specifying `.colorScheme(.light)` does not help.

<details>

```
Fatal error: _ColorSchemeKey must have a renderer-provided default value: file TokamakCore/ColorScheme.swift, line 22
Swift._assertionFailure(_: Swift.StaticString, _: Swift.String, file: Swift.StaticString, line: Swift.UInt, flags: Swift.UInt32) -> Swift.Never
static TokamakCore._ColorSchemeKey.defaultValue.getter : TokamakCore.ColorScheme
protocol witness for static TokamakCore.EnvironmentKey.defaultValue.getter : A.Value in conformance TokamakCore._ColorSchemeKey : TokamakCore.EnvironmentKey in TokamakCore
TokamakCore.EnvironmentValues.subscript.getter : <A where A: TokamakCore.EnvironmentKey>(A.Type) -> A.Value
TokamakCore.EnvironmentValues.colorScheme.getter : TokamakCore.ColorScheme
TokamakCore._SystemColorBox.resolve(in: TokamakCore.EnvironmentValues) -> TokamakCore.AnyColorBox._RGBA
TokamakCore._ColorProxy.resolve(in: TokamakCore.EnvironmentValues) -> TokamakCore.AnyColorBox._RGBA
(extension in TokamakStaticHTML):TokamakCore.Color.cssValue(TokamakCore.EnvironmentValues) -> Swift.String
(extension in TokamakStaticHTML):TokamakCore._StrokedShape.attributes(TokamakCore.ShapeStyle) -> [TokamakStaticHTML.HTMLAttribute : Swift.String]
protocol witness for TokamakStaticHTML.ShapeAttributes.attributes(TokamakCore.ShapeStyle) -> [TokamakStaticHTML.HTMLAttribute : Swift.String] in conformance TokamakCore._StrokedShape<A> : TokamakStaticHTML.ShapeAttributes in TokamakStaticHTML
(extension in TokamakStaticHTML):TokamakCore._ShapeView.deferredBody.getter : TokamakCore.AnyView
protocol witness for TokamakCore.ViewDeferredToRenderer.deferredBody.getter : TokamakCore.AnyView in conformance TokamakCore._ShapeView<A, B> : TokamakCore.ViewDeferredToRenderer in TokamakStaticHTML
[...]

PR #348, opened 2021-01-06
GTK shape support WIP

mortenbekditlevsen:
SECOND UPDATE:

* The empty environment issue is the same as #322, so I removed the workaround and added a comment to that issue.

The final thing on my todo for this PR is to test the `Path` size changes in `TokamakStaticHTML`.

UPDATE:

I marked the PR as ready for review.
There are a few issues that I am uncertain of:
* I am getting an empty environment in the `_ShapeView`, so I have hacked around that for now, which is of course not very good.
* Are the `GSignal` changes sound? Do we just keep adding overloads of `connect` for each new signal handler shape?
* Is my solution of removing `_SubPath` acceptable? (I think it makes sense).
* Is my `CGDK` systemPackage definition ok? It appears to do the trick (on macOS at least).
* I have only made brief tests on TokamakStaticHTML, but I could add an ellipse to a path and rotate it a bit (and in this case it is turned into a .path() Storage) - and this drew nicely, so something is working.

I hope to get time to install virtualbox and play around with everything on Ubuntu too.

---

Adding support for `Shape` in GTK using `gtk_drawing_area`.

Modified `Path.swift` from TokamakCore to support a few more path modifications.
I also had to remove the `_SubPath` type in the process since I could not find a way for this to work together with path modification.
The code proposed in this PR behaves (as far as I can tell) like SwiftUI in the following respects:
Adding path elements to an existing path will convert the storage to be the generic `.path` case.
Adding path storage building blocks to an empty path just sets the storage.
For instance:
adding a rectangle to an empty path will result in a path with .rectangle storage.

The current implementation of `PathBox` is just a class containing an array of `Element`... I wonder what the intention of the type is in SwiftUI - I'd expect that a `CGPath` or an array of `Element` directly as the associated value would suffice.

Still lacking is:
.trimmed and .stroked are incomplete.
In SwiftUI it's possible to get the stroked path and the trimmed path (in fact it appears that the `.description` returns a description of the 'rendered' path).
Adding stroking and trimming is very, very hard. Stroking requires the possibility of offsetting a quadratic and cubic curve, which there is not a single correct way to do.
Trimming requires measuring the length of curves, which also turns out to be difficult (but can be done by approximating a curve with line segments) as well as splitting curves.

So basically the implementation requires a lot of what is already built in to `CGPath`.
I guess that it can be added eventually by getting inspiration from something like Cairo or other open source drawing APIs.

Since the current 'backends' can stroke paths, that part isn't a huge issue, but stroking a path twice would require the functionality.

Another issue with the current implementation is that in SwiftUI, a Shape can draw outside of the frame of the shape, but that is not currently possible with the gtk version.
UPDATE: a shape can now draw out of bounds - implemented by resetting the clipping area of the widget in the drawing routine. This could probably mess up some graphics updates. I will look for a better solution once I get started on mimicking SwiftUI view sizing and layout.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
-LINKER_FLAGS := $(shell pkg-config --libs gtk+-3.0)
-C_FLAGS := $(shell pkg-config --cflags gtk+-3.0)
+LINKER_FLAGS := $(shell pkg-config --libs gtk+-3.0 gdk-3.0 pangocairo pangoft2 pango gio-unix-2.0 glib-2.0)
+C_FLAGS := $(shell pkg-config --cflags gtk+-3.0 gdk-3.0 pangocairo pangoft2 pango gio-unix-2.0 glib-2.0)
 SWIFT_LINKER_FLAGS ?= -Xlinker $(shell echo $(LINKER_FLAGS) | sed -e "s/ / -Xlinker /g" | sed -e "s/-Xlinker -Wl,-framework,/-Xlinker -framework -Xlinker /g")
 SWIFT_C_FLAGS ?= -Xcc $(shell echo $(C_FLAGS) | sed -e "s/ / -Xcc /g")
 
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
-LINKER_FLAGS := $(shell pkg-config --libs gtk+-3.0 gdk-3.0 pangocairo pangoft2 pango gio-unix-2.0 glib-2.0)
-C_FLAGS := $(shell pkg-config --cflags gtk+-3.0 gdk-3.0 pangocairo pangoft2 pango gio-unix-2.0 glib-2.0)
+LINKER_FLAGS := $(shell pkg-config --libs gtk+-3.0 gdk-3.0)
+C_FLAGS := $(shell pkg-config --cflags gtk+-3.0)
 SWIFT_LINKER_FLAGS ?= -Xlinker $(shell echo $(LINKER_FLAGS) | sed -e "s/ / -Xlinker /g" | sed -e "s/-Xlinker -Wl,-framework,/-Xlinker -framework -Xlinker /g")
 SWIFT_C_FLAGS ?= -Xcc $(shell echo $(C_FLAGS) | sed -e "s/ / -Xcc /g")
 
--- a/Package.swift
+++ b/Package.swift
@@ -91,7 +91,13 @@ let package = Package(
       pkgConfig: "gdk-3.0 pangocairo pangoft2 pango gio-unix-2.0 glib-2.0",
       providers: [
         .brew(["gtk+3", "glib", "glib-networking", "gobject-introspection"]),
-        .apt(["libgtk-3-dev", "libglib2.0-dev", "glib-networking", "gobject-introspection", "libgirepository1.0-dev"])
+        .apt([
+              "libgtk-3-dev",
+              "libglib2.0-dev",
+              "glib-networking",
+              "gobject-introspection",
+              "libgirepository1.0-dev"
+        ])
       ]),
     .target(
       name: "TokamakGTKCHelpers",
--- a/Package.swift
+++ b/Package.swift
@@ -86,13 +86,20 @@ let package = Package(
         .brew(["gtk+3"]),
       ]
     ),
+    .systemLibrary(
+      name: "CGDK",
+      pkgConfig: "gdk-3.0 pangocairo pangoft2 pango gio-unix-2.0 glib-2.0",
+      providers: [
+        .brew(["gtk+3", "glib", "glib-networking", "gobject-introspection"]),
+        .apt(["libgtk-3-dev", "libglib2.0-dev", "glib-networking", "gobject-introspection", "libgirepository1.0-dev"])
+      ]),
     .target(
       name: "TokamakGTKCHelpers",
       dependencies: ["CGTK"]
     ),
     .target(
       name: "TokamakGTK",
-      dependencies: ["TokamakCore", "CGTK", "TokamakGTKCHelpers", "CombineShim"]
+      dependencies: ["TokamakCore", "CGTK", "CGDK", "TokamakGTKCHelpers", "CombineShim"]
     ),
     .target(
       name: "TokamakGTKDemo",
--- a/Package.swift
+++ b/Package.swift
@@ -88,17 +88,13 @@ let package = Package(
     ),
     .systemLibrary(
       name: "CGDK",
-      pkgConfig: "gdk-3.0 pangocairo pangoft2 pango gio-unix-2.0 glib-2.0",
+      pkgConfig: "gdk-3.0",
       providers: [
-        .brew(["gtk+3", "glib", "glib-networking", "gobject-introspection"]),
-        .apt([
-              "libgtk-3-dev",
-              "libglib2.0-dev",
-              "glib-networking",
-              "gobject-introspection",
-              "libgirepository1.0-dev"
-        ])
-      ]),
+        .apt(["libgtk+-3.0", "gtk+-3.0"]),
+        // .yum(["gtk3-devel"]),
+        .brew(["gtk+3"]),
+      ]
+    ),
     .target(
       name: "TokamakGTKCHelpers",
       dependencies: ["CGTK"]
--- a/Sources/TokamakCore/Shapes/Path.swift
+++ b/Sources/TokamakCore/Shapes/Path.swift
@@ -59,7 +59,7 @@ public struct Path: Equatable, LosslessStringConvertible {
     case empty
     case rect(CGRect)
     case ellipse(CGRect)
-    case roundedRect(FixedRoundedRect)
+    indirect case roundedRect(FixedRoundedRect)
     indirect case stroked(StrokedPath)
     indirect case trimmed(TrimmedPath)
     case path(PathBox)
--- a/Sources/TokamakCore/Shapes/Path.swift
+++ b/Sources/TokamakCore/Shapes/Path.swift
@@ -28,14 +28,16 @@ public struct Path: Equatable, LosslessStringConvertible {
   public class _PathBox: Equatable {
     var elements: [Element] = []
     public static func == (lhs: Path._PathBox, rhs: Path._PathBox) -> Bool {
-      return lhs.elements == rhs.elements
+      lhs.elements == rhs.elements
     }
-    init() {
-    }
+
+    init() {}
+
     init(elements: [Element]) {
       self.elements = elements
     }
   }
+
   public var description: String {
     var pathString = [String]()
     for element in elements {
@@ -162,8 +164,10 @@ public struct Path: Equatable, LosslessStringConvertible {
       let minY = yPos.first ?? 0
       let maxY = yPos.last ?? 0
 
-      return CGRect(origin: CGPoint(x: minX, y: minY),
-                    size: CGSize(width: maxX - minX, height: maxY - minY))
+      return CGRect(
+        origin: CGPoint(x: minX, y: minY),
+        size: CGSize(width: maxX - minX, height: maxY - minY)
+      )
     }
   }
 
@@ -198,16 +202,16 @@ public extension Path.Storage {
     switch self {
     case .empty:
       return []
-    case .rect(let rect):
+    case let .rect(rect):
       return [
         .move(to: rect.origin),
         .line(to: CGPoint(x: rect.size.width, y: 0).offset(by: rect.origin)),
         .line(to: CGPoint(x: rect.size.width, y: rect.size.height).offset(by: rect.origin)),
         .line(to: CGPoint(x: 0, y: rect.size.height).offset(by: rect.origin)),
-        .closeSubpath
+        .closeSubpath,
       ]
 
-    case .ellipse(let rect):
+    case let .ellipse(rect):
       // Scale down from a circle of max(width, height) in order to limit
       // precision loss. Scaling up from a unit circle also looked alright,
       // but scaling down is likely a bit better.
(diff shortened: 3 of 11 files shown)


mortenbekditlevsen:
The GSignal changes are just a quick hack to get a signal connected with two arguments in the callback. Will fix.

Sources/TokamakGTK/Shapes/Shape.swift
```diff
@@ -0,0 +1,233 @@
...
+
+      let c = (style as? Color) ?? foregroundColor ?? Color.black
+//      dump(c)
+
+      // XXX TODO, don't know why my environment doesn't work.
+      var environment = EnvironmentValues()
+      environment[_ColorSchemeKey] = .light
+
+      dump(environment)
+      let rgba = _ColorProxy(c).resolve(in: environment)
```
mortenbekditlevsen, line 66: I got a crash trying to resolve the color in the environment of the _ShapeView (since that environment is empty), so I hacked around it. 
I am a bit uncertain about why I am not getting the 'actual' environment with the color scheme already set, but getting an empty one instead...

Sources/TokamakGTK/GSignal.swift
```diff
@@ -76,6 +77,47 @@ extension UnsafeMutablePointer where Pointee == GtkWidget {
     })
   }
 
+  /// Connect with a c function pointer.
+  @discardableResult
+  func connectA(
```
mortenbekditlevsen, line 82: A hack - will fix.

Sources/TokamakGTK/GSignal.swift
```diff
@@ -76,6 +77,47 @@ extension UnsafeMutablePointer where Pointee == GtkWidget {
...
+      handler,
+      data,
+      destroy,
+      GConnectFlags(rawValue: 0)
+    ))
+  }
+
+  /// Connect with a context-capturing closure (with the GtkWidget passed through)
+  @discardableResult
+  func connect2(
```
mortenbekditlevsen, line 102: A hack - will fix.

Package.swift
```diff
@@ -86,13 +86,20 @@ let package = Package(
         .brew(["gtk+3"]),
       ]
     ),
+    .systemLibrary(
+      name: "CGDK",
+      pkgConfig: "gdk-3.0 pangocairo pangoft2 pango gio-unix-2.0 glib-2.0",
```
mortenbekditlevsen, line 91: I am uncertain if all these packages are needed, I took inspiration from https://github.com/rhx/SwiftGtk

Makefile
```diff
@@ -1,5 +1,5 @@
-LINKER_FLAGS := $(shell pkg-config --libs gtk+-3.0)
-C_FLAGS := $(shell pkg-config --cflags gtk+-3.0)
+LINKER_FLAGS := $(shell pkg-config --libs gtk+-3.0 gdk-3.0 pangocairo pangoft2 pango gio-unix-2.0 glib-2.0)
```
mortenbekditlevsen, line 1: I'm not sure if all these packages are needed. I took inspiration from https://github.com/rhx/SwiftGtk

Sources/TokamakGTK/Shapes/Shape.swift
```diff
@@ -0,0 +1,153 @@
...
+// limitations under the License.
+//
+//  Created by Morten Bek Ditlevsen on 29/12/2020.
+//
+
+import CGTK
+import CGDK
+import TokamakCore
+
+func createPath(from elements: [Path.Element], in cr: OpaquePointer) {
```
mortenbekditlevsen, line 22: Perhaps this can be reused for clipping. I have an idea for a fun clipping hack:
It appears that if you connect a 'draw' signal handler, it will be called before the widget code gets a chance to draw.
This means that is should be possible to implement shape clipping by modifying the cairo context to add extra clipping for any kind of widget.
I assume that the context is 'restored' after drawing a child widget, so I expect the extra clipping to be automatically be cleaned up... Just guessing by reading docs though.
Also I think that some widgets may be drawing into their own context and not just reusing the shared window context... Will have to read up and experiment...

Sources/TokamakCore/Shapes/Path.swift
```diff
@@ -25,6 +25,17 @@ import WASILibc
 
 /// The outline of a 2D shape.
 public struct Path: Equatable, LosslessStringConvertible {
+  public class PathBox: Equatable {
```
MaxDesiatov, line 28: Can this stay `internal` as it's `internal` in SwiftUI? Or, if you have to access it in `TokamakGTK`, maybe let's keep it `public`, but rename to `_PathBox` to highlight that it's not supposed to be re-exported and exposed to users?

```suggestion
  internal class PathBox: Equatable {
```

Sources/TokamakCore/Shapes/Path.swift
```diff
@@ -139,6 +143,27 @@ public struct Path: Equatable, LosslessStringConvertible {
     case let .roundedRect(fixedRoundedRect): return fixedRoundedRect.rect
     case let .stroked(strokedPath): return strokedPath.path.boundingRect
     case let .trimmed(trimmedPath): return trimmedPath.path.boundingRect
+    case let .path(pathBox):
+      // Note: Copied from TokamakStaticHTML/Shapes/Path.swift
```
MaxDesiatov, line 147: Is this copied from that file or actually moved? Is the overall duplication reduced then?

Sources/TokamakGTK/Shapes/Shape.swift
```diff
@@ -0,0 +1,153 @@
...
+      } else {
+        elements = path.elements
+        stroke = false
+      }
+
+      if fillStyle.isEOFilled {
+        cairo_set_fill_rule(cr, cairo_fill_rule_t(rawValue: 1) /* CAIRO_FILL_RULE_EVEN_ODD */)
+      } else {
+        // This is already the default
+//        cairo_set_fill_rule(cr, cairo_fill_rule_t(rawValue: 0) /* CAIRO_FILL_RULE_WINDING */)
```
MaxDesiatov, line 109: Do we want to keep this commented?

Sources/TokamakCore/Shapes/Path.swift
mortenbekditlevsen: Excellent, I'll rename it.

Sources/TokamakCore/Shapes/Path.swift
mortenbekditlevsen: I made a single change. I'll take a look and see if I can get a better understanding of the existing logic and try to reuse the same logic both places.

Sources/TokamakGTK/Shapes/Shape.swift
mortenbekditlevsen: No, I should make it like the enum options for the strike styles. That reads quite a bit better. I'll update it.

Sources/TokamakCore/Shapes/Path.swift
mortenbekditlevsen: I _think_ that there was an error in the previous sizing and that it's fixed in the `Path.boundingRect` function, but I am not 100% certain.

The previous `elementsSize` in StaticHTML/Shapes/Path.swift had:
```
    return CGSize(width: abs(maxX - min(0, minX)), height: abs(maxY - min(0, minY)))
```
but I replaced that with
```
    return CGSize(width: maxX - minX, height: maxY - minY)
```
because otherwise the size depends on the origin, which I guess it shouldn't - unless the usage of the size needs this special behavior.
But the similar special casing was previously not applied for .rect storage type - here the size was just the rect's size (no matter the origin) - so the change seems more consistent.

@carson-katri Do you remember a use case for the special case for the previous element path sizing?

MaxDesiatov: BTW, all our renderer-independent code in TokamakCore is easy to test and there's already a test target for it, which runs only on macOS for now. I'll make it run on Ubuntu too in a future PR. Thus if you think you've found a bug in something like this, please provide a test case. Then we can test the new code against it, if that makes sense 🙂

carson-katri: I don't remember, sorry 😞

Sources/TokamakGTK/Shapes/Shape.swift
```diff
@@ -0,0 +1,158 @@
...
+      let height = gtk_widget_get_allocated_height(widget)
+
+      let c = (style as? Color) ?? foregroundColor ?? Color.black
+
+      // XXX TODO, don't know why my environment doesn't work.
+      var environment = EnvironmentValues()
+      environment[_ColorSchemeKey] = .light
+
+      let rgba = _ColorProxy(c).resolve(in: environment)
+      var color = GdkRGBA(red: Double(rgba.red),
```
carson-katri, line 77: Could this be made an extension on `Color` so we can easily get a `GdkRGBA` value from a resolved `Color` in other places too?

carson-katri (commented):
Looks great overall 👍

Sources/TokamakGTK/Shapes/Shape.swift
mortenbekditlevsen: Ah, yes - of course! Thanks!

Sources/TokamakGTK/Shapes/Shape.swift
mortenbekditlevsen: I debugged it so far to conclude that it only happens when applying stroke. So something about environment updating of a _ShapeView inside a _StrokedView?
I believe this also happens in the HTML/DOM version. At least I get a blank page when trying to stroke a shape.
I haven't really grokked all the ways an environment can be set and updated yet, so I am a bit lost here. :-/

Sources/TokamakCore/Shapes/Path.swift
mortenbekditlevsen: Hi @MaxDesiatov ,
The code was in the TokamakStaticHTML target, and I guess the best way to test would be to get a visual representation of various shapes with positive and negative origins - and compare them to SwiftUI.
I will have a look at this when I get a chance.

Sources/TokamakGTK/Shapes/Shape.swift
mortenbekditlevsen: Done.

Sources/TokamakGTK/Shapes/Shape.swift
mortenbekditlevsen: @MaxDesiatov Perhaps I could remove my workaround and create an issue for the bug - since (as far as I can tell) the issue is not related to TokamakGTK or this PR?

MaxDesiatov: Yes please, that makes perfect sense 👍

mortenbekditlevsen: It turns out that the issue is already represented as: https://github.com/TokamakUI/Tokamak/issues/322
I'll add a bit of extra information to the issue.

Sources/TokamakCore/Shapes/Path.swift
```diff
@@ -168,33 +193,222 @@ public enum RoundedCornerStyle: Hashable, Equatable {
...
+      if rect.size.width > rect.size.height {
+        transform = CGAffineTransform(scaleX: 1,
+                                      y: rect.size.height / rect.size.width)
+      } else if rect.size.height > rect.size.width {
+        transform = CGAffineTransform(scaleX: rect.size.width / rect.size.height,
+                                      y: 1)
+      } else {
+        transform = .identity
+      }
+      let elements =  [
```
MaxDesiatov, line 226: Nit: formatting is slightly off here. Could you run `swiftformat .` in the root directory with the latest SwiftFormat installed? You can also install a pre-commit hook [as described in `README.md`](https://github.com/TokamakUI/Tokamak#coding-style) to auto-format and lint all code before you commit.


```suggestion
      let elements = [
```

Sources/TokamakGTK/GSignal.swift
```diff
@@ -100,3 +146,9 @@ final class SingleParamClosureBox<T, U> {
 
   init(_ closure: @escaping (T) -> U) { self.closure = closure }
 }
+
+final class DualParamClosureBox<T, U, V> {
+  let closure: (T, U) -> V
+
+  init(_ closure: @escaping (T, U) -> V) { self.closure = closure }
```
MaxDesiatov, line 153: It's a shame that Swift doesn't support variadic generics, or macros, or something of that nature that could simplify this boilerplate. I guess something like GYB than it's worth, but we could try it later you want 🙂

MaxDesiatov (commented):
Overall LGTM, this just needs a SwiftFormat pass 🙂

Sources/TokamakGTK/GSignal.swift
mortenbekditlevsen: Right, there are quite a few good use cases for variadic generics. I hope that we'll see them one day.
I haven't looked too much into the various events/signals in gtk, but my guess is that they don't have callbacks with a whole lot of arguments... So until we see some more of them I guess this is acceptable.

Sources/TokamakCore/Shapes/Path.swift
mortenbekditlevsen: I ran swiftformat manually for now. For some reason I get additional changed files when running in the pre-commit handler - even though it's using the same swiftformat configuration. Perhaps it's because my directory is not clean.

Sources/TokamakCore/Shapes/Path.swift
mortenbekditlevsen: Ok, I think I can gather the point of the old code. I cannot figure out what's best/correct though.

The whole thing has to do with sizing of `Shape`. In SwiftUI a view is being proposed a size from it's parent - a `Shape` always accepts the proposed size.
If 'no size' is proposed (e.g. by using .fixedSize() modifier), a `Shape` uses the fallback of a 10 by 10 frame.

In other words - SwiftUI has no 'intrinsic' size for a `Shape`.
In `TokamakCore` there is added to the `Shape` a concept of fixed and flexible sizing. This concept does not appear to be part of SwiftUI (where the sizing is always determined as a function of the proposed size).
And in `TokamakStaticHTML` an intrinsic size for a `Shape` (or rather, a `Path`) is being calculated - and for the .fixed size concept, this is where the `size` calculation was used.

So previously it appeared that the calculation was basically:
The bounding rect of the `Path` 'unioned' with .zero (or something to that effect). Meaning that a bounding rect with origin 10, 10 and size 10, 10 would generate the size 20, 20.
If the path was entirely in the negative axises I think that it would be cropped away since the calculated size is always in positive.

Currently the boundingRect size is being returned, which I guess is not much better.

But the question is: what is actually the best thing to do?

SwiftUI has a different way of sizing and if there is such a thing as an intrinsic size, it appears to be 10 by 10 (which is the size of a `Shape` when applying .fixedSize()).

mortenbekditlevsen: Another comment about the `_Sizing` concept:
Right now built-in shapes like `Rectangle` are .flexible in size while custom Shapes are .fixed.
A custom shape may as well be flexible - it all depends on whether or not the `func path(in rect:)` returns a path that is relative to the rect or not...

I don't know if the _Sizing concept could be made obsolete somehow by doing something else entirely...

MaxDesiatov: I consider bringing the layout stuff in a separate PR from https://github.com/objcio/S01E235-swiftui-layout-explained-layout-priorities, would that be related in any way?

MaxDesiatov (approved):
Splendid! 👏

Sources/TokamakCore/Shapes/Path.swift
mortenbekditlevsen: I appear to get ok results by removing the `Path._Sizing` enum entirely and choosing the code paths that corresponded to the previous 'flexible' sizing...

That removes the need for the fixed size entirely - and I get quite similar results with SwiftUI and Tokamak(DOM) drawing shapes with fixed sizes - both when I create them using a 'Shape' or a 'Path' directly.

Skærmbillede 2021-01-12 kl  10 52 19
Skærmbillede 2021-01-12 kl  10 52 07

I am still uncertain if I may be lacking some context, but at least I suspect that this could be a way to go about it...

MaxDesiatov: If fixing this requires some substantial changes, I suggest doing it in a separate PR to make changes more granular and easier to track and review.

mortenbekditlevsen: Excellent. I’ll leave the sizing as it is for now. I don’t think it’s substantially more correct or incorrect than the previous version.

And regarding your question above which I just saw now:
Yes, my knowledge about the layout system indeed comes from the reverse engineering work of objc.io Swift Talks. 

I really hope to get time to look into that for GTK.
I believe that the full layout system requires you to be able to measure view sizes before they are rendered - I think that can be done in gtk, but I am not familiar with how it might be done in HTML/JavaScript although there appears to be some hacks available.

PR #355, opened 2021-01-17
Add support for shadow modifier

### LINKED ISSUE #324: Shadow support
```swift
public func shadow(
  color: Color = Color(.sRGBLinear, white: 0, opacity: 0.33),
  radius: CGFloat,
  x: CGFloat = 0, y: CGFloat = 0
) -> some View
```

> Adds a shadow to this view.
>
> The example below a series shows of boxes with increasing degrees of shadow ranging from 0 (no shadow) to 5 points of shadow, offset down and to the right of the views:
> 
> ```swift
> struct Shadow: View {
>     var body: some View {
>         HStack {
>             ForEach(0..<6) {
>                 Color.red.frame(width: 60, height: 60, alignment: .center)
>                     .overlay(Text("\($0)"),
>                              alignment: .bottom)
>                     .shadow(color: Color.gray,
>                             radius: 1.0,
>                             x: CGFloat($0),
>                             y: CGFloat($0))
>             }
>         }
>     }
> }
> ```
> 
> - Parameters:
>   - color: The shadow's color.
>   - radius: The shadow's size.
>   - x: A horizontal offset you use to position the shadow relative to
>     this view.
>   - y: A vertical offset you use to position the shadow relative to this
>     view.
> - Returns: A view that adds a shadow to this view.

literalpie:
Closes #324

### DIFF
--- a/Sources/TokamakStaticHTML/Modifiers/LayoutModifiers.swift
+++ b/Sources/TokamakStaticHTML/Modifiers/LayoutModifiers.swift
@@ -101,3 +101,11 @@ extension _PaddingLayout: DOMViewModifier {
       .joined(separator: " ")]
   }
 }
+
+extension _ShadowLayout: DOMViewModifier {
+  public var attributes: [HTMLAttribute: String] {
+    ["style": "box-shadow: \(x)px \(y)px \(radius)px 0px \(color.cssValue(.defaultEnvironment));"]
+  }
+
+  public var isOrderDependent: Bool { true }
+}
--- a/Sources/TokamakStaticHTML/Modifiers/LayoutModifiers.swift
+++ b/Sources/TokamakStaticHTML/Modifiers/LayoutModifiers.swift
@@ -104,9 +104,7 @@ extension _PaddingLayout: DOMViewModifier {
 
 extension _ShadowLayout: DOMViewModifier {
   public var attributes: [HTMLAttribute: String] {
-    [
-      "style": "box-shadow: \(x)px \(y)px \(radius * 2)px 0px \(color.cssValue(.defaultEnvironment));",
-    ]
+    ["style": "box-shadow: \(x)px \(y)px \(radius * 2)px 0px \(color.cssValue(environment));"]
   }
 
   public var isOrderDependent: Bool { true }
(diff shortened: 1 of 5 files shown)


Sources/TokamakStaticHTML/Modifiers/LayoutModifiers.swift
```diff
@@ -101,3 +101,11 @@ extension _PaddingLayout: DOMViewModifier {
       .joined(separator: " ")]
   }
 }
+
+extension _ShadowLayout: DOMViewModifier {
+  public var attributes: [HTMLAttribute: String] {
+    ["style": "box-shadow: \(x)px \(y)px \(radius)px 0px \(color.cssValue(.defaultEnvironment));"]
```
carson-katri, line 107: Maybe you can multiply the radius by 2? Seems like when going from [Sketch to UIKit](https://sarunw.com/posts/match-view-shadow-to-sketch-shadow/#blur-vs.-shadow-radius) (which uses blur and spread), they'll divide by two, so the inverse could maybe work here?

literalpie: Yes, that's at least closer than what I had (color meter says it's not a match, but it's close enough for my naked eye)

Sources/TokamakStaticHTML/Modifiers/LayoutModifiers.swift
```diff
@@ -101,3 +101,13 @@ extension _PaddingLayout: DOMViewModifier {
       .joined(separator: " ")]
   }
 }
+
+extension _ShadowLayout: DOMViewModifier {
+  public var attributes: [HTMLAttribute: String] {
+    [
+      "style": "box-shadow: \(x)px \(y)px \(radius * 2)px 0px \(color.cssValue(.defaultEnvironment));",
```
literalpie, line 108: I almost forgot about this `defaultEnvironment`. I'm not sure how to make this come from the app's actual environment.

carson-katri: I think you can add `@Environment(\.self) var environmentValues` to the modifier to access it.

MaxDesiatov:
@literalpie to fix the macOS build you'll have to open the demo Xcode project (in the `NativeDemo` subdirectory) and add the new `ShadowDemo` file to it.

Sources/TokamakStaticHTML/Modifiers/LayoutModifiers.swift
literalpie: I wasn't able to use `@Environment(\.self)`. I used `EnvironmentReader` which is how this is done in other modifiers (specifically `_BackgroundModifier`).

MaxDesiatov (approved):
Awesome, LGTM 👍

carson-katri (approved):

PR #356, opened 2021-01-18
Fixed a small issue with re-renderers being dropped

foscomputerservices:
The code was clearing the queuedRerenders set after processing all of the updates on each re-renderer.  However, during processing it is possible that new values were enqueued.  By clearing the set afterwards, these newly queued re-renderers were accidentally dropped.

This would cause some Views to not update when @State was changed.


MaxDesiatov (approved):
Great catch, thanks for the fix! 👏

MaxDesiatov:
@mortenbekditlevsen I wonder if this may fix issues you saw with GTK on Linux not re-rendering the view tree on updates?

PR #358, opened 2021-01-18
Removed an extra space that cause Safari to issue "Invalid value"

foscomputerservices:
An extra space was added at the beginning of the 'ry' attribute value of a rect.  Safari reported: Error: Invalid value for <rect> attribute ry=" 25.0".


MaxDesiatov (approved):
👏

PR #359, opened 2021-01-19
Two infinite loop fixes

foscomputerservices:
Both of these issues are fixes to `CustomStringConvertible` implementations that either directly or indirectly called themselves via `String(describing:)`.

I'm not sure what is expected to be presented in these `description` implementations, so please review and let me know if you would like alternative implementations.


MaxDesiatov (approved):
Fabulous! I guess this fixes the infinite loop you previously mentioned?

foscomputerservices:
> Fabulous! I guess this fixes the infinite loop you previously mentioned?

Yup!

PR #360, opened 2021-01-19
Use `macos-latest` agent for the GTK build

MaxDesiatov:
`macos-latest` still points to macOS 10.15 right now, which has more agent's capacity, and macOS 11.0 agents are slow and unreliable. At the same time, we have to keep using macOS 11.0 for the core build to have access to the new `App` lifecycle in the native demo project.


MaxDesiatov:
I'm merging this to make a few other PRs go faster, let me know if you'd like this to be reverted.

PR #361, opened 2021-01-19
Added TextField support for GTK using GtkEntry

mortenbekditlevsen:
Note, that in the 'update' block I do not disconnect and reconnect the signal as it is done in the gtk `Button` implementation.
I am a bit uncertain whether or not that is actually needed in some special case.
If it's not needed, perhaps it can be removed from the `Button` implementation too.


MaxDesiatov (approved):
Seems legit 👍

PR #365, opened 2021-01-20
Add a benchmark target and a script to run it

MaxDesiatov:
Benchmarks need to be built in the release mode, I created a separate `benchmark.sh` script to build it and run it.

I've also cleaned up a compiler warning in `TextEditor.swift` and bumped macOS agents to use Xcode 12.3 instead of 12.2

After this is merged to `main` I want to set it up an [similarly to what we have in JavaScriptKit](https://github.com/swiftwasm/JavaScriptKit/blob/main/.github/workflows/perf.yml) where benchmarks run on both a PR branch and `main` and difference between them is posted as a comment.


carson-katri (dismissed):

MaxDesiatov:
Just a trivial change to fix the buld, so I'm assuming that the previous review approval still holds 🙂

Issue #367, opened 2021-01-21
Infinite loops w/ 100% CPU usage caused by stack overflows


MaxDesiatov:
**Describe the bug**
There's a certain snippet of code, which when rendererd in the browser triggers an infinite loop after clicking on a `NavigationLink`.

**To Reproduce**
Build and run this snippet with `carton dev`:
```swift
import TokamakShim

struct TokamakApp: App {
  var body: some Scene {
    WindowGroup("Spooky Hanger") {
      NavigationView {
        List {
          ForEach(["Section"], id: \.self) { section in
            Section(header: Text(section).font(.headline)) {
              ForEach(["Item 1"], id: \.self) { childRow in
                NavigationLink(
                  destination: Text(childRow).padding([.leading, .trailing])
                ) {
                  Text(childRow)
                }
                .padding(.leading)
              }
            }
          }
        }
        .listStyle(SidebarListStyle())
      }
    }
  }
}

TokamakApp.main()
```
Open it in any browser a tap on "Item 1". Sometimes it loads the destination, but frequently it just hangs. If it doesn't hang, reload the page and try again. It's usually reproducible at least one in five attempts. I've reproduced in Safari, Firefox, and Chrome.

**Expected behavior**
`NavigationLink` always opens the destination without any problems.

**Screenshots**
<img width="1001" alt="Screenshot 2021-01-21 at 20 22 35" src="https://user-images.githubusercontent.com/112310/105408078-6a7cab80-5c26-11eb-92e7-e0f5eb0d2738.png">

**Additional context**

This was reported by @foscomputerservices in a bigger project, but I've managed to reduce it to this snippet. I've tried to remove seemingly unrelated modifiers or views, like `listStyle` or `padding`, but removing them seems to make the issue go away or reduce the probability or reproduction steps to something I can't trigger anymore.

MaxDesiatov:
Occasionally (but not always!) I'm able to catch an OOB with this in Chrome:

<img width="1315" alt="Screenshot 2021-01-21 at 21 17 43" src="https://user-images.githubusercontent.com/112310/105413566-2f7e7600-5c2e-11eb-924d-a3f655a3f772.png">

MaxDesiatov:
[...]

PR #370, opened 2021-01-22
Replace uses of the Runtime library with stdlib

MaxDesiatov:
This should allow us to remove the Runtime dependency eventually, which seems to be unstable, especially across different platforms and Swift versions.

Seems to resolve in one instance https://github.com/TokamakUI/Tokamak/issues/367. There are a few other places where `typeInfo` is still used, I'll clean that up in a follow-up PR.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -120,6 +120,13 @@ let package = Package(
       ]
     ),
     .target(
+      name: "TokamakCoreBenchmark",
+      dependencies: [
+        "Benchmark",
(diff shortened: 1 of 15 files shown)


carson-katri (dismissed):

MaxDesiatov:
This seems to resolve at least one instance of the bug. I'm merging it as is to make it easier to work with different branches that resolve other cases separately.

PR #371, opened 2021-01-23
Refactor environment injection, add a test

### LINKED ISSUE #367: Infinite loops w/ 100% CPU usage caused by stack overflows

MaxDesiatov:
We had some issues in this code area previously, and I refactored it a bit it in attempt to fix #367. Would be great to increase the test coverage here before further refactoring.

### DIFF
--- a/Sources/TokamakCore/MountedViews/MountedElement.swift
+++ b/Sources/TokamakCore/MountedViews/MountedElement.swift
@@ -34,6 +34,7 @@ enum MountedElementKind {
 
 public class MountedElement<R: Renderer> {
   private var element: MountedElementKind
+  var type: Any.Type { element.type }
 
   public internal(set) var app: _AnyApp {
     get {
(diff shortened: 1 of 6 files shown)


foscomputerservices (approved):
I have tested these changes against my larger application and they do fix the problem that I was having.

PR #372, opened 2021-01-24
Clean up metadata reflection code

MaxDesiatov:
Our OpenCombine fork no longer depends on Runtime, and we don't need much from it other than struct metadata. I removed the unused bits and bobs and kept only a minimal subset of it that we really need. This should make it easier for us to test and debug, as #367 has shown that some weird stuff may still lurk in that area.


foscomputerservices (approved):
I tested this branch with my code and it works! Thanks Max!

MaxDesiatov:
Perfect, please let me know if you see any similar issues. If there's anything left to do on this front, I'd like to get to the root cause of it. But so far it seems that messing with metadata through mutable pointers was to blame.

PR #375, opened 2021-01-28
Add checks for metadata state

MaxDesiatov:
Type metadata actually *is* mutable, is updated by the runtime, and we need to be careful when reading from it. See https://github.com/apple/swift/blob/main/docs/ABI/TypeMetadata.rst#metadata-requests-and-responses for more.


foscomputerservices (approved):
Looks good. Thanks @MaxDesiatov!

Issue #350, opened 2021-01-08
Set up code coverage reports on GitHub Actions


MaxDesiatov:
I think stability of our existing code should be of higher priority. We need to be diligent about keeping test coverage high and not reducing it significantly (ideally increasing it, as it's probably already too low) in any PRs. GitHub Actions job that runs in a PR and posts current coverage as a comment would be super helpful.

After that is done we can set up more complex workflows that compute coverage deltas. I know that proprietary solutions like [Codecov](codecov.io) can already do this, but I'm wary of relying on them as they tend to shut down free access for open-source projects after some period of initial unsustainable growth. GitHub Actions seems to be a more reliable tool for this long term.

mattpolzin:
In the “easy to integrate, easy to move on from” category, I can say my codecov action has been good enough for me so far (and obviously improvements are welcome!).

https://github.com/mattpolzin/swift-codecov-action

here it is in action; it currently just prints a coverage table, optionally fails if coverage standards are not met, and results in a CI checkmark on PRs. 
https://github.com/mattpolzin/OpenAPIKit/runs/1623070311?check_suite_focus=true

MaxDesiatov:
Lovely! I wonder how hard it would be to modify it to post coverage results as comments? We previously had some success in creating GitHub releases from shell scripts directly using `curl`, maybe that could be done in your action too?

mattpolzin:
One option would be to “fix” the current limitations in what my action outputs and pair it with other GitHub actions (I’m sure there’s one out there that does nothing but post a comment).

mattpolzin:
(I’m not opposed to adding features like that to my action, I just generally find GitHub actions to be so conveniently composable)

carson-katri:
Is it possible to put it in the diff like the linter comments? That way it doesn't clutter the discussion.

mattpolzin:
Definitely (assuming I catch your drift). Creating a file and committing it is not so different effort-wise to creating a new PR comment from a GitHub action.

MaxDesiatov:
Our SwiftLint warnings are attached in diffs to specific lines, while code coverage percentage is global per project.

I like the idea of committing that percentage as a file though, that way it's simpler to compare coverage of a PR branch with coverage in the main branch. We wouldn't have to run tests again to calculate the value for the main branch then. We could just check out the main branch in the job and load the value from the file.

MaxDesiatov:
[...]

PR #378, opened 2021-02-01
Simple Code Coverage analysis

mattpolzin:
Closes https://github.com/TokamakUI/Tokamak/issues/350.

Adds simple code coverage analysis. Until the GitHub token is set up in this repo, you can see the results including a comment on the PR here: https://github.com/mattpolzin/Tokamak/pull/2

I only commented out the tests to make the coverage diff with `main` interesting. Obviously not expecting that change to get merged :)


MaxDesiatov:
This is awesome, thanks! I wonder if coverage diffs could be posted as a GitHub Action logging command? Then it would require any tokens for comments. It would look similar to this, visible only in diffs of PRs:

screenshot

[According to the GitHub Actions docs](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-a-warning-message) it's enough to output a line in this form to stdout on a build agent for it to be displayed this way: `::warning file=Package.swift,line=1,col=1::Code coverage summary goes here...`. [The SwiftLint action](https://github.com/norio-nomura/action-swiftlint/blob/9f4dcd7fd46b4e75d7935cf2f4df406d5cae3684/entrypoint.sh#L14) formats its output in a similar way.

What do you think about this approach?

mattpolzin:
@MaxDesiatov That's an interesting idea. Makes me think it would be _really_ cool to tag the appropriate Swift file's line 1 when that file has low test coverage (or has decreased test coverage when compared to main) but that's just an idea, not something easy to support without a bit more work on the underlying coverage analysis tool.

RE your idea to log the results against the Package file, that's worth a try.  I wonder if it will syntax highlight the diff or if we want to just print the table in there without coloring (or ditch the diff idea and just print the current coverage table?). 

I'd rather have a nicer diff table, anyway, with a single `Color.swift     -19.34%` line instead of the two lines indicating `19.34%` was removed and `0%` was added -- the current diffing is just a limitation of doing this with what I've got now rather than implementing a new feature on the underlying tool.

To wrap up this message which is more about things I wish for the future than it is about the present, I can play with this more (try out the logging approach) in a few days (or possibly next weekend if I get busy). At the moment I've reached my limit for dealing with how frustrating it can be to troubleshoot random GitHub Action issues :)

MaxDesiatov:
No problem at all, I think just logging as is against `Package.swift` (even the first line) is a good enough approach. Thanks for your hard work!

mattpolzin:
@MaxDesiatov that seems to work; I see the annotation in the Files Changed tab at the bottom.

Would you want to always print the codecov as a warning this way or only log the warning if the coverage was too low. I suppose it could also be a warning normally and an error if the coverage was too low (which is going to fail that CI step anyway)

MaxDesiatov:
> I suppose it could also be a warning normally and an error if the coverage was too low (which is going to fail that CI step anyway)

Yes, this makes perfect sense. Even if the coverage is improved it's convenient to observe how exactly it changed 👍

mattpolzin:
Well, that seems to be working. On my draft PR linked in this PRs description you can see a failing result whereas this PR passes.

Final question: What do you want the minimum coverage to be initially?

If you like this where it's at, I'll uncomment those tests to get it ready for merging.

MaxDesiatov:
As far as I understand, 20% is the current setting, LGTM. Thanks!

MaxDesiatov (approved):
Amazing work! 👏

j-f1:
This is a OK start, but would it be possible to [integrate codecov](https://github.com/codecov/example-swift) so people can get a direct link to view more detailed coverage info?

MaxDesiatov:
We've used codecov.io in the past, before SwiftUI was announced and Tokamak had a React-like API, but I found it to be not very stable overall and hard (if possible at all) to debug if coverage reports somehow don't appear in their UI. As it is a paid and apparently a VC-funded service, I'm also afraid it will eventually stop serving open-source projects like Travis did.

Maybe we could upload coverage reports to S3 or GitHub Pages instead? (In some next iteration of this)

mattpolzin:
There might be a way to use GitHub Action artifacts to expose a public URL for the full test coverage table (percent per file).

The action already produces `./codecov.txt`. It's easy to upload that file as an artifact, but there won't be a direct link to it by name -- instead, we would need to query the GitHub rest API for artifacts related to the current workflow run and then take the right artifact's ID and plug it into `/repos/{owner}/{repo}/actions/artifacts/{artifact_id}` ourselves to produce a direct link.

MaxDesiatov:
Thanks for taking care of this! I think it's reasonable for more advanced features to be added in separate PRs.

PR #382, opened 2021-02-17
Make properties of CGPoint, CGSize and CGRect vars instead of lets

mortenbekditlevsen:
And initialized array used by CGAffineTransform concatenation

### DIFF
--- a/Sources/TokamakCore/Stubs/CGStubs.swift
+++ b/Sources/TokamakCore/Stubs/CGStubs.swift
@@ -25,8 +25,8 @@ import WASILibc
 
 public typealias CGFloat = Double
 public struct CGPoint: Equatable {
-  public let x: CGFloat
-  public let y: CGFloat
+  public var x: CGFloat
+  public var y: CGFloat
 
   public init(x: CGFloat, y: CGFloat) {
     self.x = x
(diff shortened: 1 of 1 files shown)


mortenbekditlevsen:
Note:
The `rotated(by:)` implementation appears to be a copy of the `init(rotationAngle:)` - it doesn't take the current matrix values into account at all.
Would it suffice to return `self.concatenating(Self(rotationAngle: angle))` ?

MaxDesiatov (approved):
Long term we should use Foundation types directly instead of our own stubs, but for now this seems legit.

carson-katri:
We should probably separate it out to it's own product and do
```swift
#if canImport(CoreGraphics)
@_exported import CoreGraphics
#else
...
#endif
```
since I doubt CG will be available on the web, for instance.

MaxDesiatov:
CoreGraphics itself is not available, but all of the types and some of their extensions in `CGStubs` are already available in Foundation for WASI. I've tried using that instead of our own reimplementation, but that broke some of our `Shape` SVG rendering. We need to add unit tests to thoroughly cover that area and make the switch. We already link with Foundation for `Bundle` required by the `Image` type anyway.

PR #383, opened 2021-02-21
Stop assuming pointer type for generic arguments

MaxDesiatov:
Seems to fully resolve https://github.com/TokamakUI/Tokamak/issues/367.

You were right @foscomputerservices, this was a strict aliasing issue. 😅 This [forum post](https://forums.swift.org/t/is-it-safe-to-use-a-variable-after-modifying-via-a-rebound-pointer/44848/6) made it crystal clear to me:

>Binding a local variable's memory to a different type poisons that local variable

### DIFF
--- a/Sources/TokamakCore/Reflection/Layouts/FieldDescriptor.swift
+++ b/Sources/TokamakCore/Reflection/Layouts/FieldDescriptor.swift
@@ -54,14 +54,14 @@ extension UnsafePointer where Pointee == FieldRecord {
 
   func type(
     genericContext: UnsafeRawPointer?,
-    genericArguments: UnsafeRawPointer?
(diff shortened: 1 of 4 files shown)


carson-katri (approved):
Nice!

MaxDesiatov:
This doesn't make the issue go away. I'm still seeing it in slightly larger apps, and I still can't reproduce it with a test renderer to add a unit test. I suspect this only shuffles some memory around and makes the set of conditions for reproduction shift. Working on a slightly different approach for this right now...

MaxDesiatov:
Even though we've found a different way to solve the original issue, I'd like to keep this open (albeit in the draft state) until we have some benchmarking automation to compare if there are any performance or code size benefits by any chance.

PR #384, opened 2021-02-23
Reuse reflection code from the Swift toolchain

### LINKED ISSUE #367: Infinite loops w/ 100% CPU usage caused by stack overflows

MaxDesiatov:
Yet another attempt to resolve #367. ~Seems to work in all known cases so far.~

This reuses code that OpenCombine previously used for reflection, but didn't need after all. It's transplanted from the toolchain with namespace names changed to avoid naming collisions.


MaxDesiatov:
I'd like to keep this open (albeit in the draft state) until we have some benchmarking automation to compare if there are any performance or code size benefits by any chance.

carson-katri:
Could the benchmarks we have now open this back up?

MaxDesiatov:
Yes, they could. I'm not sure when I'll have time to do set up these benchmarks though, sorry Feel free to pick it up if you're interested.

MaxDesiatov:
Closed by mistake while commenting, reopening now...

PR #394, opened 2021-03-26
Add `DatePicker` to the `TokamakDOM` module

### LINKED ISSUE #320: Implement `DatePicker` view in the DOM renderer
Relevant API:

```swift
public struct DatePickerComponents : OptionSet {
  public let rawValue: UInt
  public init(rawValue: UInt)
  public static let hourAndMinute: DatePickerComponents
  public static let date: DatePickerComponents
  public typealias Element = DatePickerComponents
  public typealias RawValue = UInt
  public typealias ArrayLiteralElement = DatePickerComponents
}

public struct DatePicker<Label> : View where Label : View {
  public typealias Components = DatePickerComponents
  public var body: some View {
    get
  }
}

public protocol DatePickerStyle {
  associatedtype _Body : View
  func _body(configuration: DatePicker<Self._Label>) -> Self._Body
  typealias _Label = _DatePickerStyleLabel
}

public struct _DatePickerStyleLabel : View {
  public typealias Body = Swift.Never
}

extension View {
  public func datePickerStyle<S>(_ style: S) -> some View where S : DatePickerStyle
}
```

Snowy1803:
This fixes #320 by adding a SwiftUI-compatible `DatePicker`. However, `DatePickerStyle` is not supported. 

This uses the HTML inputs `date`, `time`, or `datetime-local`, depending on the given `displayedComponents`. This means that not all browsers show the picker, as Mac Safari currently does not support them. Safari on Mac will just show an ISO-format text field. If the date is in an invalid format, the binding will not receive updates until it becomes parseable by JSDate.

On supported browsers, the binding gets updated in real time, as you would expect, with a Foundation.Date, just like SwiftUI.

### DIFF
--- a/Sources/TokamakCore/Views/Selectors/DatePicker.swift
+++ b/Sources/TokamakCore/Views/Selectors/DatePicker.swift
@@ -1,0 +1,174 @@
+// Copyright 2020 Tokamak contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import struct Foundation.Date
+
+/// A control for selecting a value from a bounded linear range of values.
+///
+/// Available when `Label` and `ValueLabel` conform to `View`.
+public struct DatePicker<Label>: PrimitiveView where Label: View {
+  let label: Label
+  let valueBinding: Binding<Date>
+  let displayedComponents: DatePickerComponents
+  let min: Date?
+  let max: Date?
+
+  public typealias Components = DatePickerComponents
+}
+
+public extension DatePicker {
+  init(
+    selection: Binding<Date>,
+    in range: ClosedRange<Date>,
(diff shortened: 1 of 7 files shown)


MaxDesiatov (commented):
Would you be able to also add something like `DatePickerDemo` to the `TokamakDemo` target? This would make the new view more discoverable alongside other already implemented views.

Snowy1803:
Sure. I found timezone issues while testing the demo, I'll fix that

Sources/TokamakCore/Views/Selectors/DatePicker.swift
```diff
@@ -0,0 +1,177 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//  Created by Emil Pedersen on 2021-03-26.
+//
+
+import struct Foundation.Date
+
+/// A control for selecting a value from a bounded linear range of values.
```
carson-katri, line 20: This seems like a description of a Slider. Maybe try
```suggestion
/// A control for selecting a `Date`.
```
or
```suggestion
/// A control for selecting a `Date` from a bounded linear range of dates.
```

carson-katri (commented):
Overall LGTM — Thanks!

Snowy1803:
Didn't notice I left that comment. Replaced it with the SwiftUI overview.

MaxDesiatov:
Please also add `DatePickerDemo.swift` to `NativeDemo/TokamakDemo.xcodeproj` project so that the macOS build passes. Thanks!

Snowy1803:
Done. Also got a warning that `*.xcodeproj` is gitignored

MaxDesiatov (approved):
Awesome, thanks!

PR #397, opened 2021-04-09
Add @inlinable where SwiftUI has it

carson-katri:
SwiftUI marks a lot of inits and View extension funcs `@inlinable`, so I went in and did the same. Not sure if it will lead to any performance differences. Do you all think this worth it/something we should do?


MaxDesiatov (approved):
It will be hard to say what impact it has on performance until we have working benchmarks integrated with CI. But I don't see any harm in it, in fact it may make calculating diffs with the SwiftUI API easier. Either visually, or in an automated way, since attributes would match too.

aehlke:
this would be nice to merge

PR #408, opened 2021-05-30
Replace `ViewDeferredToRenderer`, fix renderer tests

MaxDesiatov:
This allows writing tests for `TokamakStaticHTML`, `TokamakDOM`, and `TokamakGTK` targets.

The issue was caused by conflicting `ViewDeferredToRenderer` conformances declared in different modules, including the `TokamakTestRenderer` module. 

This works around a general limitation in Swift, which was [discussed at length on Swift Forums previously](https://forums.swift.org/t/an-implementation-model-for-rational-protocol-conformance-behavior/37171). When multiple conflicting conformances to the same protocol (`ViewDeferredToRenderer` in our case) exist in different modules, only one of them is available in a given binary (even a test binary). Also, only of them can be loaded and used. Which one exactly is loaded can't be known at compile-time, which is hard to debug and leads to breaking tests that cover code in different renderers. We had to disable `TokamakStaticHTMLTests` for this reason.

The workaround is to declare two new functions in the `Renderer` protocol:

```swift
public protocol Renderer: AnyObject {
  // ...
  // Functions unrelated to the issue at hand skipped for brevity.

  /** Returns a body of a given pritimive view, or `nil` if `view` is not a primitive view for
   this renderer.
   */
  func body(for view: Any) -> AnyView?

  /** Returns `true` if a given view type is a primitive view that should be deferred to this
   renderer.
   */
  func isPrimitiveView(_ type: Any.Type) -> Bool
}
```

Now each renderer can declare their own protocols for their primitive views, i.e. `HTMLPrimitive`, `DOMPrimitive`, `GTKPrimitive` etc, delegating to them from the implementations of `body(for view:)` and `isPrimitiveView(_:)`. Conformances to these protocols can't conflict across different modules. Also, these protocols can have `internal` visibility, as opposed to `ViewDeferredToRenderer`, which had to be declared as `public` in `TokamakCore` to be visible in renderer modules.


Sources/TokamakCore/MountedViews/MountedApp.swift
```diff
@@ -30,7 +30,7 @@ final class MountedApp<R: Renderer>: MountedCompositeElement<R> {
     // They also have no parents, so the `parent` argument is discarded as well.
     let childBody = reconciler.render(mountedApp: self)
 
-    let child: MountedElement<R> = mountChild(childBody)
+    let child: MountedElement<R> = mountChild(reconciler.renderer, childBody)
```
MaxDesiatov, line 33: `renderer` instances now need to be explicitly pass through in a lot more places. This allows mounted elements to call back to the renderer to determine whether a given view is primitive for this renderer.

Sources/TokamakCore/StackReconciler.swift
```diff
@@ -201,45 +202,50 @@ public final class StackReconciler<R: Renderer> {
...
 
-    return compositeElement[keyPath: result](compositeElement[keyPath: bodyKeypath])
+    return compositeElement[keyPath: keyPath]
   }
 
   func render(compositeView: MountedCompositeView<R>) -> AnyView {
-    render(compositeElement: compositeView, body: \.view.view, result: \.view.bodyClosure)
+    let view = body(of: compositeView, keyPath: \.view.view)
+
+    guard let renderedBody = renderer.body(for: view) else {
```
MaxDesiatov, line 236: I've only refactored `ViewDeferredToRenderer` for now. `SceneDeferredToRenderer` is kept as is to keep the PR small. And we didn't have problems with scenes (yet).

Sources/TokamakCore/Views/AnyView.swift
```diff
@@ -50,21 +50,8 @@ public struct AnyView: PrimitiveView {
...
-          }
-          // swiftlint:disable:next force_cast
-          return (deferredView as! ViewDeferredToRenderer).deferredBody
-        }
-      } else {
-        // swiftlint:disable:next force_cast
-        bodyClosure = { AnyView(($0 as! V).body) }
-      }
+      // swiftlint:disable:next force_cast
+      bodyClosure = { AnyView(($0 as! V).body) }
```
MaxDesiatov, line 54: So happy this mess in `AnyView` can be finally simplified!

Sources/TokamakCore/Views/Buttons/Button.swift
```diff
@@ -48,7 +48,7 @@ public struct Button<Label>: View where Label: View {
   }
 }
 
-public struct _Button<Label>: PrimitiveView where Label: View {
+public struct _Button<Label>: _PrimitiveView where Label: View {
```
MaxDesiatov, line 51: `PrimitiveView` is not meant to be re-exported by any renderer modules, making that explicit here with underscore naming.

Tests/TokamakStaticHTMLTests/HTMLTests.swift
```diff
@@ -18,6 +18,149 @@
 import TokamakStaticHTML
 import XCTest
 
+private let expectedHTML =
```
MaxDesiatov, line 21: It would be nice to use https://github.com/pointfreeco/swift-snapshot-testing here instead, but we're running these tests in the Wasm environment, and file access is tricky there. Hardcoding the test fixture here for now.

Sources/TokamakCore/Renderer.swift
```diff
@@ -66,4 +66,14 @@ public protocol Renderer: AnyObject {
     with host: MountedHost,
     completion: @escaping () -> ()
   )
+
+  /** Returns a body of a given pritimive view, or `nil` if `view` is not a primitive view for
+   this renderer.
+   */
+  func body(for view: Any) -> AnyView?
```
carson-katri, line 73: Would it make sense for this to be called `primitiveBody(for:)`? Then it’s more clear that it only returns the body of a primitive view?

carson-katri (commented):
Looks good to me, excited to get some good testing going.

Issue #403, opened 2021-05-14
Adding padding to a view contained in a Button causes the Button to disappear


foscomputerservices:
When adding padding to the content of a ```Button```, the ```<button>``` is no longer rendered in the HTML.

```swift
struct ContentView: View {
    var body: some View {
        Button {
        } label: {
            Text("Button")
                .padding()
        }
    }
}
```

Removing the ```.padding()``` restores the button in the output.

**Screenshots**
(I could provide a screenshot, but just envision a blank screen, we'll save a few bytes on GitHub 😁)

This is using the latest code on the *main* branch.

MaxDesiatov:
Closing this as resolved in #409.

PR #409, opened 2021-06-07
Allow DOMRenderer to render buttons with non-Text labels (#403)

ezraberch:
Currently, `DOMRenderer` can only handle `Button`s where is the label is `Text`. If it is any other `View`, the `Button` is not rendered. This is the cause of #403.

This PR removes this restriction. Additionally, it expands the `ButtonStyle` demo to include `Button`s with complex labels.


MaxDesiatov (approved):
LGTM, thanks!

Issue #395, opened 2021-03-27
Content with spacer jumps when blurring and focusing the page 


revolter:
**Describe the bug**

When using a combination of `ZStack`, `VStack` and `Spacer`, some views jump when changing to another tab and back again.

**To Reproduce**
1. Download the attached project
2. Run `carton dev`
3. Open the webpage
4. Select another tab
5. Select the initial tab

**Expected behavior**

The text should remain outside the rectangle.

**Screenshots**
If this is a layout/rendering issue, please provide screenshots for both Tokamak and SwiftUI that highlight the difference.

**Desktop (please complete the following information):**
 - OS: macOS Big Sur 11.2.3 (20D91)
 - Browser: Safari
 - Version of the browser: 14.0.3 (16610.4.3.1.7)
 - Version of Tokamak: 0.6.1

**Smartphone (please complete the following information):**
 - Version of the browser [e.g. 22]
 - Version of Tokamak [e.g. 0.6.1]

**Additional context**

[1616847509.zip](https://github.com/TokamakUI/Tokamak/files/6215857/1616847509.zip)

MaxDesiatov:
Thanks for the detailed report and sorry for the delayed reply! I assume that we can fix this by our layout code rewrite in https://github.com/TokamakUI/Tokamak/pull/379, but there may be an easier fix. Unfortunately, I don't have an estimate for when I'll be able to investigate this in a detailed way. In the meantime, I'm happy to review PRs that fix this.

PR #410, opened 2021-06-15
Fix spacers after DOMRenderer.update

ezraberch:
`DOMRenderer.mount` contains code which can be necessary to properly render spacers. However, `update` can overwrite what this code does, leading to the problem described in #395.

This PR modifies `update` to fix this issue.

### DIFF
--- a/Sources/TokamakDOM/DOMRenderer.swift
+++ b/Sources/TokamakDOM/DOMRenderer.swift
@@ -95,6 +95,16 @@ final class DOMRenderer: Renderer {
     ) { scheduler.schedule(options: nil, $0) }
   }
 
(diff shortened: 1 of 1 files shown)


carson-katri:
This looks good to me.

I had an idea for a different way to implement Spacers though, which we could possibly use now that we have PreferenceKey.

Basically we'd have two Spacer views, `_HSpacer`, and `_VSpacer` (SwiftUI actually has these views), and Spacer would swap its body out depending on an environment variable passed down by VStack/HStack. Then the spacer views could send a preference up to the V/HStack specifying that they exist in this context. They could also pass that up to their parent, so it requests flexible width/height.

MaxDesiatov (approved):
Seems legit, thanks!

PR #411, opened 2021-06-16
Use `CGFloat`, `CGPoint`, `CGRect` from Foundation

### LINKED ISSUE #404: Use `NSGeometry` types from Foundation
**Describe the bug**
`CGRect` and other related types created in Tokamak aren't the same as Foundation types and can't be passed to functions/types expecting Foundation types.

We should be using types declared in https://github.com/apple/swift-corelibs-foundation/blob/eec4b26deee34edb7664ddd9c1222492a399d122/Sources/Foundation/NSGeometry.swift instead of rolling our own versions. I've tried that at some point, but that broke our SVG rendering. We need to cover that with tests and do a safe migration to the original Foundation's version.

MaxDesiatov:
Resolves #404.

This also allows us to write more tests that are source-compatible with SwiftUI.

### DIFF
--- a/Sources/TokamakCore/Modifiers/Effects/ClipEffect.swift
+++ b/Sources/TokamakCore/Modifiers/Effects/ClipEffect.swift
@@ -1,4 +1,4 @@
-// Copyright 2020 Tokamak contributors
+// Copyright 2020-2021 Tokamak contributors
 //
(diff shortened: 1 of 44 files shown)


foscomputerservices (approved):
Looks great to me ✅

PR #412, opened 2021-06-19
Add a snapshot test for `Path` SVG layout

MaxDesiatov:
Depends on #253.

This adds a dependency on the [SnapshotTesting](https://github.com/pointfreeco/swift-snapshot-testing) library, which allows testing our SVG layout algorithm end-to-end. We use the `--screenshot` flag of Chromium-based browsers (MS Edge in this case) to produce a PNG snapshot of a view rendered with `StaticHTMLRenderer`.

This test works only on macOS for now due to its dependency on `NSImage`, but that should be fine as we'd expect the same SVG output to be rendered in the same way on all platforms.


Sources/TokamakCore/Shapes/Path/PathLayout.swift
```diff
@@ -12,185 +12,11 @@
...
 // limitations under the License.
 //
-//  Created by Carson Katri on 06/28/2020.
+//  Created by Max Desiatov on 20/06/2021.
 //
 
 import Foundation
 
-/// The outline of a 2D shape.
-public struct Path: Equatable, LosslessStringConvertible {
```
MaxDesiatov, line 21: Moved to a separate `Path.swift` file to fix linter warnings.

Tests/TokamakTests/ReconcilerStressTests.swift
```diff
@@ -40,20 +40,9 @@ final class ReconcilerStressTests: XCTestCase {
     let renderer = TestRenderer(SpookyHanger())
     let root = renderer.rootTarget
 
-    return
```
MaxDesiatov, line 43: This test was invalid, but the `return` statement made this undiscoverable. Cleaning that up and simplifying the test to actually pass.

Sources/TokamakCore/Shapes/Path/Path.swift
```diff
@@ -0,0 +1,192 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//  Created by Carson Katri on 06/28/2020.
+//
+
+import Foundation
+
+/// The outline of a 2D shape.
+public struct Path: Equatable, LosslessStringConvertible {
```
MaxDesiatov, line 21: This is a smaller version of the original `Path.swift` that satisfies linter's file length constraints.

foscomputerservices (approved):
Nice!

PR #413, opened 2021-06-21
Explicitly import CoreFoundation

yonihemi:
This allows TokamakDemo to be built with [latest SwiftWasm 5.4 toolchain](https://github.com/swiftwasm/swift/releases/tag/swift-wasm-5.4-SNAPSHOT-2021-06-17-a) in **debug mode** (compiler still crashes when building for release).

BTW this import could be anywhere in the target, couldn't find a file that felt natural to include it in.

### DIFF
--- a/Sources/TokamakCore/PreviewProvider.swift
+++ b/Sources/TokamakCore/PreviewProvider.swift
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import Foundation
+import CoreFoundation
 
 /// This protocol has no functionality currently, and is only provided for compatibility purposes.
 public protocol PreviewProvider {
--- a/Sources/TokamakCore/PreviewProvider.swift
+++ b/Sources/TokamakCore/PreviewProvider.swift
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 import Foundation
-import CoreFoundation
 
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@kateinoigakukun do you think this is a good approach in the light of your changes to how CoreFoundation is linked?

MaxDesiatov:
To clarify, I remember Foundation folks saying that they plan to make CoreFoundation "private" and not available on non-Apple platforms, since its use there is very limited and may cause some issues in the future. Overall I'm leaning towards merging this as a temporary fix, but at first I would like to hear what Yuta thinks, if he's not too busy right now of course 🙂

kateinoigakukun:
Importing CoreFoundation is not exactly right approach, but it's ok to merge this as a temporary solution

MaxDesiatov (dismissed):
Thanks!

Sources/TokamakCore/PreviewProvider.swift
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import Foundation
+import CoreFoundation
```
MaxDesiatov, line 16: Maybe moving this to `CGStubs.swift` would be more appropriate?

MaxDesiatov (approved):

PR #414, opened 2021-06-25
Fix handling of stroked shapes

ezraberch:
The PR fixes multiple bugs which prevent stroked shapes from rendering correctly.

1. Allow environment injection into  `_StrokedShape`. This causes stroked shapes to no longer crash (#322).
2. Change `Path.strokedPath` and `Path.trimmedPath` to allow the `sizing` of the base shape to be inherited. Without this, stroked shapes can appear as 0x0 in size, making them invisible.
3. Change `_ShapeView` in the StaticHTML renderer to merge attributes in the `svg` rather than placing the `svg` in a `div`. This allows proper rendering when multiple shapes are in a stack.

Finally, the `Path` demo has been modified to add a stroked circle.


MaxDesiatov (changes requested):
Thanks for the PR! I'm seeing distortion in Firefox and Chromium-based browsers, how hard would it be to fix that?

Screenshot 2021-06-27 at 22 37 18

Also, would you be able to add a snapshot test here? https://github.com/TokamakUI/Tokamak/blob/main/Tests/TokamakStaticHTMLTests/LayoutTests.swift

ezraberch:
That distortion is actually an issue with `ZStack` unrelated to stroked shapes. You can see the same issue on `main` with a normal circle. So I've replaced the `ZStack` with an `HStack`. I've also added a snapshot test, as requested.

MaxDesiatov (approved):
Great stuff, thanks!

PR #169, opened 2020-07-07
Add `Toolbar` implementation for HTML renderer

carson-katri:
`Toolbar` is new in SwiftUI. It is coupled fairly closely with `NavigationView`, so this should be integrated with that somehow (#130). I made it similar to macOS which allows more than a leading/trailing `ToolbarItem`.

Resolves #316.

PR #415, opened 2021-07-04
Add HTML implementation for `opacity` modifier

MaxDesiatov:
Let's move this out of #169 to make that PR a bit smaller and add more tests.

Depends on #273 to avoid conflicts in the testing code.


carson-katri (approved):

Sources/TokamakStaticHTML/Modifiers/Effects/OpacityEffect.swift
```diff
@@ -0,0 +1,24 @@
...
+// limitations under the License.
+//
+//  Created by Carson Katri on 1/20/21.
+//
+
+import TokamakCore
+
+extension _OpacityEffect: DOMViewModifier {
+  public var attributes: [HTMLAttribute: String] {
+    ["style": "opacity: \(opacity)"]
```
ezraberch, line 22: `Circle().frame(width: 100, height: 100, alignment: .center).opacity(0.5)`will render the `style` improperly:

`<div style="opacity: 0.5width: 100.0px;`...

```suggestion
    ["style": "opacity: \(opacity); "]
```

MaxDesiatov: Great catch, thanks!

PR #416, opened 2021-07-07
Add ContainerRelativeShape

carson-katri:
This adds `ContainerRelativeShape` and a `containerShape` modifier. It does not correctly handle the position of Views, as `GeometryProxy` doesn't provide access to View frames, only bounds.


MaxDesiatov (commented):
Overall seems legit. 👍
Could this have a run of `swiftformat` to resolve warnings, and also a snapshot test added?

MaxDesiatov (approved):
Awesome, thanks!

PR #418, opened 2021-07-07
Add `Material` to the HTML renderer

carson-katri:
This adds the new iOS 15+ `Material` ShapeStyle. They currently only work for `background` modifiers on the web using `backdrop-filter`.

Tokamak:
Dark Mode Materials
Light Mode Materials

SwiftUI:
Dark Mode Materials
Light Mode Materials


carson-katri:
I added a snapshot test, but the filter doesn't seem to be screenshot correctly...

Tests/TokamakStaticHTMLTests/RenderingTests.swift
```diff
@@ -203,6 +203,31 @@ final class RenderingTests: XCTestCase {
     )
     #endif
   }
+
+  func testMaterial() {
+    #if compiler(>=5.5) || os(WASI)
```
MaxDesiatov, line 208: Just curious, what features of 5.5 are used here? Since we don't have the latest Xcode on GitHub Actions, these tests won't run there, right?

carson-katri: It just needs the newest SwiftUI in Xcode 13 so it can find `Material`, and some new modifiers. I wasn't sure what the correct way to restrict it to that is besides using Xcode 13 on CI or restricting to the compiler version shipping with that Xcode version.

MaxDesiatov: Right, but these tests should be using `Material` from Tokamak, not SwiftUI. There's no `import SwiftUI` in `RenderingTests.swift` at all, and there's an explicit import of `TokamakStaticHTML`, so maybe removing this `#if compiler(>=5.5)` should be an issue? And `os(WASI)` shouldn't have any effect since snapshot tests aren't supported on that platform anyway, and there's a top-level platform check for macOS in that file.

carson-katri: Oops, sorry. Mixed that up with the demo files which wouldn't build. I'll get rid of it.

MaxDesiatov (approved):
Awesome! Seems legit

PR #419, opened 2021-07-07
Add Primary/Secondary/Tertiary/QuaternaryContentStyle

carson-katri:
These styles have 2 different behaviors:
1. Use the styles specified in a `foregroundStyle` modifier:
```swift
Group {
  Text("Primary").foregroundStyle(.primary)
  Text("Secondary").foregroundStyle(.secondary)
  Text("Tertiary").foregroundStyle(.tertiary)
  Text("Quaternary").foregroundStyle(.quaternary)
}
.foregroundStyle(.red, .green, .blue)
```
Screen Shot 2021-07-07 at 5 43 14 PM

2. Use the foreground color and apply opacity:
```swift
Group {
  Text("Primary").foregroundStyle(.primary)
  Text("Secondary").foregroundStyle(.secondary)
  Text("Tertiary").foregroundStyle(.tertiary)
  Text("Quaternary").foregroundStyle(.quaternary)
}
.foregroundColor(.blue)
```
Screen Shot 2021-07-07 at 5 44 16 PM


Sources/TokamakCore/Shapes/ShapeStyles/ContentStyles.swift
```diff
@@ -0,0 +1,104 @@
...
+  @inlinable
+  public init() {}
+}
+
+extension PrimaryContentStyle: ShapeStyle {
+  public func _apply(to shape: inout _ShapeStyle_Shape) {
+    if !shape.inRecursiveStyle,
+       let foregroundStyle = shape.environment._foregroundStyle
+    {
+      if foregroundStyle.styles[0] is Self {
```
MaxDesiatov, line 28: Do we have a certainty that `foregroundStyle` is always non-empty? If so, maybe this needs a comment in code to explain that?

Also, this PR needs conflicts to be resolved 🙂

carson-katri: I replaced the array with a tuple to guarantee its non-empty.

MaxDesiatov (approved):
Sweet! 🙌

PR #420, opened 2021-07-08
Check minWidth/Height == nil

### LINKED ISSUE #387: Frame sizes do not match expected behavior.
**Describe the bug**
Defining a full width or -height view does not match expectations, unless you set the `minWidth` or `minHeight` explicitly to `0`.

**To Reproduce**
The following code fails
```swift
WindowGroup("Tokamak App") {
    Color.black.frame(
        maxWidth: .infinity,
        maxHeight: .infinity
    )
}
```

**Expected behavior**
The entire view is black.

**Screenshots**
The view is only black in a strip in the center.

image

**Desktop:**
 - OS: **macOS**
 - Browser **Safari**
 - Version of the browser **14.0.3 (16610.4.3.1.4)**
 - Version of Tokamak **0.6.1**

**Additional context**
It's simply fixed by modifying the `fillWidth` and `fillHeight` checks from `minWidth == 0` to `minWidth ?? 0 == 0`. I'd also suggest a similar change be made when no frame is set, to implicitly make the view infinite width/height, unless there's an exception such as `Text` or `TextField` which are bound by font size.

carson-katri:
Resolves #387 

This also updates frame layouts to respect the `alignment` parameter.


carson-katri:
I couldn't see any visual difference in the opacity test, so I replaced the image. If you do see a difference please let me know and I'l try to resolve it.

MaxDesiatov (approved):
Nice! 👏

PR #422, opened 2021-07-08
Add `aspectRatio` modifier

carson-katri:
This adds the `aspectRatio`, `scaledToFill`, and `scaledToFit` modifiers and the `ContentMode` enum.

I also added the `resizable` modifier to `Image`, and made it respect the specified `aspectRatio`.


Sources/TokamakCore/Modifiers/AspectRatioLayout.swift
```diff
@@ -0,0 +1,70 @@
...
+  case fill
+}
+
+public struct _AspectRatioLayout: ViewModifier {
+  public let aspectRatio: CGFloat?
+  public let contentMode: ContentMode
+
+  @inlinable
+  public init(aspectRatio: CGFloat?, contentMode: ContentMode) {
+    (self.aspectRatio, self.contentMode) = (aspectRatio, contentMode)
```
MaxDesiatov, line 28: nit: from coding style perspective I'd prefer it to be on separate lines, just for consistency's sake with the rest of the codebase.

Sources/TokamakCore/Views/Image.swift
```diff
@@ -17,27 +17,121 @@
...
-  public init(decorative name: String, bundle: Bundle? = nil) {
-    label = nil
-    self.name = name
-    self.bundle = bundle
+  override func resolve(in environment: EnvironmentValues) -> ResolvedValue {
+    .init(storage: .named(name, bundle: bundle), label: label)
+  }
+}
+
+private class ResizableProvider: AnyImageProviderBox {
```
MaxDesiatov, line 67: A bit unsure about this class hierarchy, are we going to have more than one subclass of `AnyImageProviderBox`?

Sources/TokamakCore/Views/Image.swift
```diff
@@ -17,27 +17,121 @@
 
 import Foundation
 
-public struct Image: _PrimitiveView {
-  let label: Text?
+public class AnyImageProviderBox: AnyTokenBox, Equatable {
```
MaxDesiatov, line 20: Should this have an underscore prefix to indicate that this shouldn't be re-exported?

```suggestion
public class _AnyImageProviderBox: AnyTokenBox, Equatable {
```

Sources/TokamakCore/Views/Image.swift
carson-katri: We will likely have something like `SystemImageProvider` in the future.

MaxDesiatov: Ah, you mean for something like SF Symbols?

carson-katri: Yeah, exactly.

carson-katri: Those will need to be resolved in the environment so they can access the font and accessibility settings, so I think it makes sense to have it be a token.

Tests/TokamakStaticHTMLTests/RenderingTests.swift
```diff
@@ -254,6 +254,18 @@ final class RenderingTests: XCTestCase {
       timeout: defaultSnapshotTimeout
     )
   }
+
+  func testAspectRatio() {
+    assertSnapshot(
+      matching: Ellipse()
+        .fill(Color.purple)
+        .aspectRatio(0.75, contentMode: .fit)
```
MaxDesiatov, line 262: The PR looks good to me otherwise, but I just noticed that you only test `.fit` here, can a case for `.fill` also be added?

MaxDesiatov (approved):
Thanks!

PR #424, opened 2021-07-09
Add `scaleEffect` modifier

carson-katri:
This uses the `scale()` CSS transform in the HTML/DOM renderers.


Sources/TokamakStaticHTML/Modifiers/Effects/ScaleEffect.swift
```diff
@@ -0,0 +1,24 @@
...
+// limitations under the License.
+//
+//  Created by Carson Katri on 7/9/21.
+//
+
+import TokamakCore
+
+extension _ScaleEffect: DOMViewModifier {
+  public var attributes: [HTMLAttribute: String] {
+    ["style": "transform: scale(\(scale.width), \(scale.height));"]
```
MaxDesiatov, line 22: Does this mean that `anchor` property has no effect in the HTML renderer?

carson-katri: Oh, yes. I'll add the `transform-origin` property.

carson-katri: I fixed the rotationEffect modifier as well.

MaxDesiatov (approved):
Awesome, thank you!

PR #425, opened 2021-07-09
Add ProgressView

carson-katri:
This adds `ProgressView` using the `progress` tag on the web.

Screen Shot 2021-07-09 at 2 02 54 PM


Sources/TokamakCore/Views/Progress/ProgressView.swift
```diff
@@ -0,0 +1,203 @@
...
+      configuration: .init(
+        fractionCompleted: fractionCompleted,
+        label: label.map { .init(body: AnyView($0)) },
+        currentValueLabel: currentValueLabel.map { .init(body: AnyView($0)) }
+      )
+    )
+  }
+}
+
+// FIXME: `Foundation.Progress` not found.
```
MaxDesiatov, line 68: Maybe this could be uncommented and `#if !os(WASI)` added around it?

MaxDesiatov (commented):
Aside from my `#if !os(WASI)` suggestion, would it be possible to add a snapshot test here? What exactly is preventing us from making it available in the HTML renderer instead of DOM?

carson-katri:
Yeah, I can move it to the static renderer 👍

MaxDesiatov (approved):
Great stuff, thanks!

MaxDesiatov:
Could you also update [the `progress.md` doc](https://github.com/TokamakUI/Tokamak/blob/main/docs/progress.md)?

carson-katri:
@MaxDesiatov Is there a way to see the failing snapshots taken by the GH action? I changed the progress doc and it failed...

MaxDesiatov:
Lemme check, it should upload failing snapshots as artefacts, but I hope that re-running it will make it go away 🙏

carson-katri:
Does it maybe need to be `RenderingTests` not `SnapshotTests`?

MaxDesiatov:
I've fixed snapshot uploads, let me know if that helps you in diagnosing. Otherwise I can have a closer look in the next couple of days.

.github/workflows/ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           # avoid building unrelated products for testing by specifying the test product explicitly
           swift build --product TokamakPackageTests
           `xcrun --find xctest` .build/debug/TokamakPackageTests.xctest ||
-            find /var/folders -iname SnapshotTests -exec cp -r {} . \;
+            cp -r `find /var/folders -iname RenderingTests` . && exit 1
```
ezraberch, line 40: This doesn't work as intended. See [this](https://unix.stackexchange.com/questions/88850/precedence-of-the-shell-logical-operators).

```suggestion
            (cp -r `find /var/folders -iname RenderingTests` . ; exit 1)
```

MaxDesiatov (approved):
Really not sure what the snapshot errors were about, glad it works after merging the `main` branch.

LGTM 🙌

PR #426, opened 2021-07-10
Add View Traits and transitions

carson-katri:
This adds `_ViewTraitKey` and related internal modifiers. Then transition builds on top of these traits.

Here's the transition demo:
https://user-images.githubusercontent.com/13581484/126044318-b2856cbb-56ed-4196-a0fd-6ea5c6c606c7.mov

It works by applying the transition's modifier to the a View on mount/update/unmount.


Sources/TokamakDOM/DOMNode.swift
```diff
@@ -54,7 +54,13 @@ extension _AnimationBoxBase._Resolved._RepeatStyle {
 }
 
 extension AnyHTML {
-  func update(dom: DOMNode, transaction: Transaction) {
+  func update(
+    dom: DOMNode,
+    additionalAttributes: [HTMLAttribute: String],
+    transaction: Transaction
+  ) {
```
MaxDesiatov, line 61: Would it be possible to split the body of this function into multiple functions or some helper types to resolve the linter warning?

Sources/TokamakDOM/Modifiers/Transitions.swift
```diff
@@ -0,0 +1,35 @@
...
+import TokamakStaticHTML
+
+extension _MoveTransition: DOMViewModifier {
+  public var attributes: [HTMLAttribute: String] {
+    let offset: (String, String)
+    switch edge {
+    case .leading: offset = ("-100%", "0")
+    case .trailing: offset = ("100%", "0")
+    case .top: offset = ("-100%", "0")
+    case .bottom: offset = ("100%", "0")
```
MaxDesiatov, line 28: Shouldn't these 4 all be different from each other? Right now `leading`/`trailing` and `top`/`bottom` seem to be using the same axis.

Sources/TokamakCore/MountedViews/MountedElement.swift
```diff
@@ -98,24 +98,29 @@ public class MountedElement<R: Renderer> {
     }
   }
 
+  public internal(set) var viewTraits: _ViewTraitStore
```
MaxDesiatov, line 101: Are these traits somehow different from SwiftUI accessibility traits? I can't find anything else about SwiftUI traits, just wondering what's used as a reference here.

carson-katri: SwiftUI has `_ViewTraitKey` internally, and uses them for `transition`, `onDelete`, `onMove` and some other modifiers (I think drag & drop?). From what I can tell, a trait is data attached to a single view, and in many cases seems to be a way to get the trait out of SwiftUI and into the host platform (for instance, for UITableView actions or NSItemProvider-based drag & drop).

Although I'm really just speculating as to how SwiftUI uses them by looking at the `swiftinterface`. In Tokamak they are meant to be passed to the nearest host element so renderers can access them.

Sources/TokamakDOM/Modifiers/Transitions.swift
carson-katri: Yes, I'll fix that 👍

Sources/TokamakCore/MountedViews/MountedHostView.swift
```diff
@@ -57,7 +60,9 @@ public final class MountedHostView<R: Renderer>: MountedElement<R> {
     guard !view.children.isEmpty else { return }
 
     mountedChildren = view.children.map {
-      $0.makeMountedView(reconciler.renderer, target, environmentValues, self)
+      let child = $0.makeMountedView(reconciler.renderer, target, environmentValues, self)
+      child.viewTraits = self.viewTraits
+      return child
```
ezraberch, line 65: Since `viewTraits` is being set immediately, it would make sense to let `makeMountedView` set it directly.

ezraberch:
1. If you remove all but one of the items in the demo (leaving `Text(".slide").transition(AnyTransition.slide)`, for example), the text will fade in and out instead of showing the proper transition.
2. For the following, the SwiftUI behavior is the first line slides while the second and third lines fade. With your code, the second and third lines both slide and fade.
```swift
Text(".slide")
  .transition(AnyTransition.slide)
VStack {
  Text(".slide").transition(AnyTransition.slide)
  Text(".slide").transition(AnyTransition.slide)
}
```
3. For the following, all 3 lines slide. However, unlike in SwiftUI, line 1 and lines 2-3 do not slide together when sliding out.
```swift
Text(".slide")
  .transition(AnyTransition.slide)
VStack {
  Text(".slide")
  Text(".slide")
}.transition(AnyTransition.slide)
```
4. If you click fast enough, you can see the same element multiple times.

Sources/TokamakDOM/DOMNode.swift
```diff
@@ -78,106 +111,90 @@ extension AnyHTML {
...
-            res[key] = computedStyle?[dynamicMember: key].string
-              ?? dom.ref.style.object?[key].string
-          } else {
-            res[key] = dom.ref.style.object?[key].string
-          }
+  private func animateStyles(
+    _ style: String,
+    on dom: DOMNode,
+    with animation: Animation
+  ) {
```
MaxDesiatov, line 121: I realize these linter warnings are a bit annoying, but is it still possible to get it a bit slimmer and under 50 lines? Maybe would work better from code readability perspective too.

carson-katri:
1-3 should be fixed.

carson-katri:
My idea to fix 4 atm is to keep a list of `unmountingViews` (Views that haven't called the `callback` after unmounting yet), and keep those available for reconciliation, then cancel their removal if they are required to mount again.

ezraberch:
> 1-3 should be fixed.

1 and 2 look good.

For 3, the slide-out looks good. However, for the slide-in, lines 2-3 just appear rather than sliding.

ezraberch:
`onDisappear` is triggering before the animation rather than after. With the following additions/changes to the Demo:
```swift
@State private var showText = "Show"
...
Button(isVisible ? "Hide" : showText) {
...
Text(".opacity")
  .transition(AnyTransition.opacity)
  .onDisappear(perform: {
    showText = "Show Again"
})
```

In SwiftUI, the button text changes to "Show" when you click "Hide", then changes to "Show Again" when the animation is complete. With your code, it changes from "Hide" to "Show Again" immediately upon clicking "Hide".

carson-katri:
> For 3, the slide-out looks good. However, for the slide-in, lines 2-3 just appear rather than sliding.

Seems like it only affects `.move` (`.slide`)...

carson-katri:
4 should be fixed now too.

carson-katri:
> `onDisappear` is triggering before the animation rather than after.

There's already a `FIXME` for both `onAppear` and `onDisappear` to be renderer-specific, so maybe we could defer that to a separate PR (using something like the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) to check if the element is visible in the viewport)?

ezraberch (changes requested):
For 3, the slide-in is still behaving incorrectly.

For 4, the fix is causing another problem. If you switch from Transition to a different section of the demo while a transition is happening, the animating HTML will appear in the new section for the remaining duration of the transition animation.

ezraberch:
Also, I'm now getting an error when rapidly pressing the inner button:

```
Fatal error: Attempted to read an unowned reference but object 0x733ee0 was already deallocated
Uncaught RuntimeError: unreachable
```

carson-katri:
The slide-in bug should be fixed. Seems like the `getComputedStyle` function returns the transformation matrix in pixel values instead of percentages, which are incorrect because the children haven't been mounted yet. That's probably why it worked for single elements but not `VStack`. I disabled computing the start keyframe when the element is mounting to work around this.

Sources/TokamakCore/MountedViews/MountedHostView.swift
```diff
@@ -56,55 +62,89 @@ public final class MountedHostView<R: Renderer>: MountedElement<R> {
...
+    )
   }
 
   override func update(in reconciler: StackReconciler<R>, with transaction: Transaction) {
     guard let target = target else { return }
 
+    // Stop any unfinished unmounts and complete them without transitions.
+    parentUnmountTask.cancel()
+    parentUnmountTask.completeImmediately()
+    parentUnmountTask = .init()
```
ezraberch, line 121: If `scenePhase` changes to `background` then back to `active` (changing tabs, for example), you get a stack buffer overflow error. This happens even from the Demo main page. Commenting out these lines removes the error.

Sources/TokamakDemo/TransitionDemo.swift
```diff
@@ -0,0 +1,74 @@
...
+}
+
+struct TransitionDemo: View {
+  @State private var isVisible = false
+  @State private var isInnerVisible = false
+
+  var body: some View {
+    VStack {
+      Button(isVisible ? "Hide" : "Show") {
+        withAnimation(.easeInOut(duration: 3)) { isVisible.toggle() }
```
ezraberch, line 36: If you remove the animation from the outer section (so just `isVisible.toggle()`), then click hide while the inner section is animating, the outer section will animate as well.

ezraberch (changes requested):

Sources/TokamakCore/MountedViews/MountedHostView.swift
carson-katri: This can be fixed with #367. I also removed the recursion from `UnmountTask`, maybe that will do something 🤷‍♂️

Sources/TokamakDemo/TransitionDemo.swift
carson-katri: This should be resolve now.

Sources/TokamakDemo/TransitionDemo.swift
```diff
@@ -0,0 +1,76 @@
...
+          .transition(AnyTransition.asymmetric(insertion: .opacity, removal: .scale))
+        Text(".opacity + .slide")
+          .transition(AnyTransition.opacity.combined(with: .slide))
+        Text(".modifier")
+          .transition(AnyTransition.modifier(
+            active: ColorOverlayModifier(color: .red),
+            identity: ColorOverlayModifier(color: .clear)
+          ))
+        Text(".animation")
+          .transition(AnyTransition.scale.animation(.spring()))
```
ezraberch, line 61: There's been a regression since the last batch of commits. Previously, if you remove the animation from the outer section, this (and only this) item will still animate. This matches SwiftUI behavior. With the latest changes, this item will no longer animate in this situation.

carson-katri: Root animated transitions should now run even if the transaction has no animation.

Sources/TokamakCore/MountedViews/MountedElement.swift
filip-sakel: Traits are also used for `zIndex(_:)` and `layoutPriority(_:)`. I haven't look into whether Tokamak has implemented these modifiers. If they're not implemented, traits would be really helpful.

ezraberch (dismissed):

ezraberch:
Looks good!

You are getting a couple of linter warnings now, so you may want to refactor a bit before merge.

MaxDesiatov (approved):
👏

PR #427, opened 2021-07-12
Animation implementation using the Web Animations API

carson-katri:
This builds on the stubs added in #235. I hooked up `Transaction` in the reconciler, and setup support for the `withAnimation` global function to apply animations on Binding changes.

https://user-images.githubusercontent.com/13581484/125217689-46be2800-e28f-11eb-91b4-63c5a642fddb.mov

> Currently the actual animatable properties of an Animatable View/Modifier are ignored, and all styles are animated.


Sources/TokamakCore/Animation/Animatable.swift
```diff
@@ -0,0 +1,151 @@
...
+  associatedtype AnimatableData: VectorArithmetic
+  var animatableData: Self.AnimatableData { get set }
+}
+
+public protocol _PrimitiveAnimatable {}
+
+public extension Animatable where Self: VectorArithmetic {
+  var animatableData: Self {
+    get { self }
+    set {}
```
MaxDesiatov, line 30: Is there some nice way to satisfy the linter here? Or is explicitly silencing `unused_setter_value` the only way here?

Sources/TokamakCore/Animation/Animation.swift
```diff
@@ -0,0 +1,215 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+public let defaultDuration = 0.35
```
MaxDesiatov, line 17: Would a comment here that describes how this value was inferred make sense? Or just a reassuring sentence that this was visually tested to match the duration in SwiftUI on Apple platforms?

Sources/TokamakCore/Animation/_AnimationBoxBase.swift
```diff
@@ -0,0 +1,182 @@
...
+  func equals(_ other: _AnimationBoxBase) -> Bool {
+    fatalError("implement \(#function) in subclass")
+  }
+
+  public static func == (lhs: _AnimationBoxBase, rhs: _AnimationBoxBase) -> Bool {
+    lhs.equals(rhs)
+  }
+}
+
+class StyleAnimationBox: _AnimationBoxBase {
```
MaxDesiatov, line 104: nit: would `final` make sense here?

```suggestion
final class StyleAnimationBox: _AnimationBoxBase {
```

Sources/TokamakCore/Animation/_AnimationSolvers.swift
```diff
@@ -0,0 +1,63 @@
...
+  /// Solve value at a specific point in time.
+  func solve(at t: Double) -> Double
+  /// Calculates the duration of the animation to a specific presision.
+  func restingPoint(precision y: Double) -> Double
+}
+
+public enum _AnimationSolvers {
+  /// Calculates the animation of a spring with certain properties.
+  ///
+  /// For some useful information, see [Demystifying UIKit Spring Animations](https://medium.com/ios-os-x-development/demystifying-uikit-spring-animations-2bb868446773)
```
MaxDesiatov, line 31: This line needs to be either split or linter silenced if that's not possible.

Sources/TokamakCore/Animation/_AnimationSolvers.swift
```diff
@@ -0,0 +1,63 @@
...
+      w0 = sqrt(stiffness / 2)
+      wd = sqrt(abs(pow(w0, 2) - pow(ƛ, 2)))
+      v0 = initialVelocity
+    }
+
+    public func solve(at t: Double) -> Double {
+      let y: Double
+      if ƛ < w0 {
+        y = pow(M_E, -(ƛ * t)) * ((s0 * cos(wd * t)) + ((v0 + s0) * sin(wd * t)))
+//      } else if ƛ > w0 { // Skip overdamping
```
MaxDesiatov, line 52: Is this intentionally commented? If so, can we add some clarification in the comment what's blocking it from being uncommented?

Sources/TokamakCore/Shapes/ModifiedShapes.swift
```diff
@@ -52,6 +62,20 @@ public struct _TrimmedShape<S>: Shape where S: Shape {
...
+  public typealias AnimatableData = AnimatablePair<
+    S.AnimatableData,
+    AnimatablePair<CGFloat, CGFloat>
+  >
+  public var animatableData: AnimatableData {
+    get {
+      .init(shape.animatableData, .init(startFraction, endFraction))
+    }
+    set {
+      shape.animatableData = newValue[].0
```
MaxDesiatov, line 75: Huh, I didn't know that empty subscripts are possible? Would you mind clarifying how does it actually work?

Sources/TokamakDemo/AnimationDemo.swift
```diff
@@ -0,0 +1,120 @@
...
+      Button("Start") {
+        withAnimation(Animation.default.repeatForever()) {
+          foreverAnimation = !foreverAnimation
+        }
+      }
+    }
+  }
+
+  var body: some View {
+//    Circle()
```
MaxDesiatov, line 108: Can this commented out code be cleaned up?

Sources/TokamakStaticHTML/Modifiers/LayoutModifiers.swift
```diff
@@ -131,9 +131,11 @@ extension _PaddingLayout: DOMViewModifier {
   }
 }
 
-extension _ShadowLayout: DOMViewModifier {
+extension _ShadowEffect._Resolved: DOMViewModifier {
   public var attributes: [HTMLAttribute: String] {
-    ["style": "box-shadow: \(x)px \(y)px \(radius * 2)px 0px \(color.cssValue(environment));"]
+    [
+      "style": "box-shadow: \(offset.width)px \(offset.height)px \(radius * 2)px 0px \(color.cssValue);",
```
MaxDesiatov, line 137: nit: line length linter warning

Sources/TokamakStaticHTML/Tokens/Tokens.swift
```diff
@@ -16,8 +16,13 @@ import TokamakCore
...
-    let rgba = _ColorProxy(self).resolve(in: environment)
-    return "rgba(\(rgba.red * 255), \(rgba.green * 255), \(rgba.blue * 255), \(rgba.opacity))"
+    _ColorProxy(self).resolve(in: environment).cssValue
+  }
+}
+
+extension AnyColorBox.ResolvedValue {
+  var cssValue: String {
+    "rgba(\(red * 255), \(green * 255), \(blue * 255), \(opacity))"
   }
```
MaxDesiatov, line 26: While we're at it, can TokamakCore/Tokens/Color.swift` file be split into multiple files to fix the file length warning?

Sources/TokamakCore/Shapes/ModifiedShapes.swift
carson-katri: The empty subscript just returns a tuple of the `(first, second)` elements of the pair. SwiftUI has it internally, but it'd work just as well as a property.

Sources/TokamakCore/Animation/Animatable.swift
carson-katri: Doesn't look like I can get around it without using `newValue`:
```swift
set { _ = newValue }
```
I think I'll just disable it manually.

MaxDesiatov (dismissed):
Seems legit, fantastic work!

carson-katri:
There doesn't seem to be an API for computing style values at a particular step which made solvers tricky. But I updated it to use solvers by creating an animation, reading the computed styles, then cancelling it for each step, and then creating keyframes from those steps. Seems to work well enough.

I also made the animations interruptible to match SwiftUI.

MaxDesiatov (approved):
LGTM!

PR #429, opened 2021-07-17
Fix background/overlay layout in DOM/HTML renderers

### LINKED ISSUE #402: .background modifier with contained shape causes view to expand to full vertical size of the screen
When a shape (say a RoundedRectangle()) is added to the background modifier, the view fills the entire vertical height of the screen.

```swift
struct ContentView: View {
    var body: some View {
        Text("Hello, world!")
            .background(RoundedRectangle(cornerRadius: 10.0).fill(Color.red))
    }
}
```

**Expected behavior**

On iOS the background modifier does not change the view size, so I expected the size to remain the size of ```Text```.

**Screenshots**
Attached

Screen Shot 2021-05-14 at 6 06 19 PM

**Workaround**

In this particular example, the following code can be used to work around the issue:

```swift
        Text("Hello, world!")
            .background(Color.red)
            .cornerRadius(10)
```

_NOTE: This issue was previously described along with many other issues in issue #336.  I wanted to provide a contrite example._

carson-katri:
Resolves #402 
Screen Shot 2021-07-17 at 1 19 52 PM


MaxDesiatov (approved):
Seems legit 👍

PR #431, opened 2021-07-18
Add controlSize/controlProminence modifiers

carson-katri:
This adds the `controlSize` and `controlProminence` modifiers, as well as styles for Button that use them.

Screen Shot 2021-07-17 at 9 54 35 PM

I used the `default-button` appearance which works in Safari for increased prominence buttons, but this isn't supported in other browsers. I can switch it to use custom CSS, but then it may look out of place with other browser buttons...


MaxDesiatov:
Would it be possible to use `default-button` only on Safari and custom CSS in other browsers?

carson-katri:
I used `@supports (-webkit-appearance: default-button)` to provide a custom style for browsers that don't support `default-button`.

Screen Shot 2021-07-18 at 9 52 03 AM

MaxDesiatov (approved):
Seems legit, thanks!

PR #435, opened 2021-07-28
Revise ShapeStyle and add Gradients

carson-katri:
This adds some small revisions to match SwiftUI in Xcode beta 4, as well as adding support for `Gradient` and related styles.

SwiftUI:
Screen Shot 2021-07-28 at 4 06 09 PM
Tokamak:
Screen Shot 2021-07-28 at 4 05 53 PM


Sources/TokamakCore/Shapes/ShapeStyles/Gradient.swift
```diff
@@ -0,0 +1,440 @@
...
+          endRadiusFraction: endRadiusFraction
+        )
+      )
+    )
+  }
+
+  public static func _apply(to type: inout _ShapeStyle_ShapeType) {}
+}
+
+@frozen public struct AngularGradient: ShapeStyle, View {
```
MaxDesiatov, line 207: Could this be split into separate `AngularGradient.swift` file to fix the linter warning? May be applicable to other types in this file as well.

Sources/TokamakStaticHTML/Shapes/_ShapeView.swift
```diff
@@ -56,26 +56,77 @@ extension _ShapeView: _HTMLPrimitive {
...
+        }
+      default:
+        if let color = resolved?.color(at: 0) {
+          attributes = ["style": "fill: \(color.cssValue(environment));"]
+        } else if let foregroundStyle = environment._foregroundStyle,
+                  let color = foregroundStyle.resolve(
+                    for: .resolveStyle(levels: 0..<1),
+                    in: environment,
+                    role: Content.role
+                  )?.color(at: 0)
```
MaxDesiatov, line 123: nit: a suggestion for more consistent indentation

```suggestion
        } else if
          let foregroundStyle = environment._foregroundStyle,
          let color = foregroundStyle.resolve(
            for: .resolveStyle(levels: 0..<1),
            in: environment,
            role: Content.role
          )?.color(at: 0)
```

Tests/TokamakStaticHTMLTests/RenderingTests.swift
```diff
@@ -205,21 +205,41 @@ final class RenderingTests: XCTestCase {
...
           .frame(width: 50, height: 50)
-          .foregroundStyle(QuaternaryContentStyle())
+          .foregroundStyle(HierarchicalShapeStyle.quaternary)
       }
       .foregroundColor(.blue),
       as: .image(size: .init(width: 275, height: 100)),
       timeout: defaultSnapshotTimeout
     )
+
+    assertSnapshot(
```
MaxDesiatov, line 224: nit: time to split `RenderingTests` into separate `XCTestCase` subclasses and files I guess

Tests/TokamakStaticHTMLTests/__Snapshots__/HTMLTests/testOptional.1.txt
```diff
@@ -160,6 +160,7 @@ height: 100%;
 
 "><svg style="width: 100%;
 height: 100%;
-overflow: visible; fill: rgba(255.0, 58.650000000000006, 48.45, 1.0);"><rect height="100%" stroke-width="0.0" width="100%" x="0.0" y="0.0"></rect></svg>
+overflow: visible; fill: rgba(255.0, 58.650000000000006, 48.45, 1.0);"><rect height="100%" stroke-width="0.0" width="100%" x="0.0" y="0.0"></rect>
+<defs></defs></svg>
```
MaxDesiatov, line 164: Does this `<defs></defs>` have any impact on rendering? If not, maybe our code could check if it's empty and avoid emitting it? IDK if that's the correct balance to strike here. I realize these additional checks may have impact on performance, but I'm for the smallest markup generated if possible. WDYT?

carson-katri: No it's just a container element. I'll add a check 👍

Sources/TokamakDemo/ShapeStyleDemo.swift
```diff
@@ -92,14 +92,44 @@ struct ShapeStyleDemo: View {
...
       }
+      VStack {
+        Text("Primary")
+          .foregroundStyle(HierarchicalShapeStyle.primary)
+        Text("Secondary")
+          .foregroundStyle(HierarchicalShapeStyle.secondary)
+        Text("Tertiary")
+          .foregroundStyle(HierarchicalShapeStyle.tertiary)
+        Text("Quaternary")
+          .foregroundStyle(HierarchicalShapeStyle.quaternary)
```
ezraberch, line 111: SwiftUI supports the following shorter form syntax, so we should as well (and use it in some of the demo).
```suggestion
        Text("Primary")
          .foregroundStyle(.primary)
        Text("Secondary")
          .foregroundStyle(.secondary)
        Text("Tertiary")
          .foregroundStyle(.tertiary)
        Text("Quaternary")
          .foregroundStyle(.quaternary)
```

carson-katri: Correct me if I’m wrong but I believe that requires Swift 5.5. The extensions on ShapeStyle to allow it are there, but for now we’ll have to be explicit about the type name. https://github.com/apple/swift-evolution/blob/main/proposals/0299-extend-generic-static-member-lookup.md

MaxDesiatov: Yes, I had the same impression that this requires 5.5, and we haven't even got SwiftWasm 5.4 ready for Tokamak yet...

ezraberch: Sure, but we can get the same functionality in other ways. It just takes a bit more code.

Here's what I'm thinking: ezraberch/Tokamak@9ea989d0c551ed07e4201a7f8501a355ed76d1c3

carson-katri: If we did work around that, it may be better to add something more general like the original SwiftUI beta's `StaticMember<Base>` struct. At the same time, I wonder if that would overlap with the `ShapeStyle` extensions which should currently work with `TokamakStaticHTML`/`TokamakGTK` in Swift 5.5.

MaxDesiatov: Either way, I'm totally fine with merging it as it is. It would be easier to discuss workarounds in a separate PR/issue, or maybe we'll get 5.5 working by that point.

ezraberch:
This isn't properly rendering in Safari:

DemoSafari

carson-katri:
That looks to be a layout issue that's not specifically related to the changes in this PR, and could be fixed in a separate one if that seems appropriate.

Tests/TokamakStaticHTMLTests/RenderingTests.swift
carson-katri: I tried to split the rendering tests into logical groupings, but let me know if you have any suggestions.

ezraberch:
Sure, but we should change the demo to avoid that issue for now, which can be done easily.

Also note: the material section being on the right side rather than centered is another similar issue.

Sources/TokamakDemo/ShapeStyleDemo.swift
```diff
@@ -92,14 +92,44 @@ struct ShapeStyleDemo: View {
...
+        )
+      Rectangle()
+        .fill(
+          RadialGradient(
+            colors: [.red, .green, .blue],
+            center: .topLeading,
+            startRadius: 50,
+            endRadius: 100
+          )
+        )
```
ezraberch, line 132: Explicitly set frame to avoid layout issue on Safari.

```suggestion
      Rectangle()
        .fill(
          LinearGradient(
            colors: [.red, .green, .blue],
            startPoint: .bottomLeading,
            endPoint: .topTrailing
          )
        )
        .frame(width: 300, height: 150)
      Rectangle()
        .fill(
          RadialGradient(
            colors: [.red, .green, .blue],
            center: .topLeading,
            startRadius: 50,
            endRadius: 100
          )
        )
        .frame(width: 300, height: 150)
```

Tests/TokamakStaticHTMLTests/RenderingTests/Snapshotting+image.swift
```diff
@@ -0,0 +1,71 @@
+// Copyright 2021 Tokamak contributors
```
MaxDesiatov, line 1: nit: can this be renamed to `Snapshotting+Image.swift` for consistency with capitalization?

Sources/TokamakCore/Tokens/Color/Color.swift
```diff
@@ -130,6 +130,20 @@ public extension Color {
   }
 }
 
+public extension ShapeStyle where Self == Color {
```
ezraberch, line 133: Could you include the new colors from the beta here (brown, cyan, indigo, mint, teal)?

Sources/TokamakStaticHTML/Shapes/_ShapeView.swift
```diff
@@ -56,26 +56,78 @@ extension _ShapeView: _HTMLPrimitive {
...
+      case let .gradient(gradient, style):
+        let stops = ForEach(Array(gradient.stops.enumerated()), id: \.offset) {
+          HTML("stop", [
+            "offset": "\($0.element.location * 100)%",
+            "stop-color": $0.element.color.cssValue(environment),
+          ])
+        }
+        let id = Int.random(in: 0..<Int.max)
+        attributes = ["style": "fill: url(#gradient\(id));"]
+        switch style {
```
ezraberch, line 80: Looks like only linear and radial gradients are implemented here. Are you not planning to implement the others for this PR?

When you add the others, I'd suggest refactoring this `switch` to reduce code duplication.

carson-katri: I don't think SVG supports the other gradient types, but I could look for a workaround.

ezraberch: CSS gradients look like they would work:

https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/conic-gradient()

carson-katri: Can you use CSS gradient in SVG?

ezraberch: You can put the SVG in a clipPath, then use a foreignObject containing the CSS gradient with the clipPath.

See [here](https://css-tricks.com/my-struggle-to-use-and-animate-a-conic-gradient-in-svg/) for an example.

Would this approach work for our purposes?

MaxDesiatov: I'd really prefer other gradient types to be split into separate PRs. Ideally we'd want to keep PRs as small as possible, which makes it easier to review them and to go through history if needed.

MaxDesiatov (approved):
Seems legit 👍

Issue #423, opened 2021-07-09
Sanitation of User Input


ezraberch:
Many web applications allow for custom content shared between users, such as user comments. Failure to sanitize that content can lead to security issues. #421 adds some basic sanitation functionality, but only in a very specific context. The problem is far broader.

To see the issue, go to the `TextField` section of the demo and paste this in the first text field:
`<a onMouseOver="alert('Arbitrary JavaScript Executed')">Some Text</a>`

Mousing over the rendered text will execute the injected JavaScript.

This can be protected against at the app level, but it is best not to rely on that, especially since this isn't an issue in a traditional SwiftUI app.

While text displayed should be sanitized by default, it is important to provide a mechanism for app developers to disable this functionality when required (for text coming from specific trusted users, for example). It would also be beneficial to define multiple sanitation levels/strategies; it might be desirable to allow things like `<b>` while still filtering out custom JavaScript.

aehlke:
DOMPurify is a good lib for this

PR #437, opened 2021-08-05
Add HTML sanitizer to Text

ezraberch:
This PR adds basic HTML sanitation to `Text`, fixing the vulnerability mentioned in #423. It does the following:

- Add a new `sanitized` option for `_Storage`. It works the same as `verbatim`, except that it indicates that the `Text` does not need further sanitization.
- Add two sanitizers. `Sanitizers.HTML.Encode` will encode control characters. This is now the default for rendering `Text` when no sanitizer is specified explicitly. `Sanitizers.HTML.Insecure` does no sanitization; this can be used to tell the renderer not to sanitize a `Text`.
- Add new constructors to `Text` to allow specification of a sanitizer, using either a built-in or custom sanitizer. Sanitizers are simple `String -> String` closures.
- Add functionality to TokamakShim so code using the new functionality will compile in non-DOM/HTML environments.
- Updated the demo and added tests.


Sources/TokamakShim/TokamakDOMAdditions.swift
```diff
@@ -0,0 +1,32 @@
...
+#if !os(WASI)
+public enum Sanitizers {
+  public enum HTML {
+    public static let insecure: (String) -> String = { $0 }
+    public static let encode: (String) -> String = { $0 }
+  }
+}
+
+public extension Text {
+  init(verbatim content: String, sanitizer: (String) -> String) { self = Self(content) }
```
carson-katri, line 27: Maybe putting the sanitizer in the `Environment` would be better for this. Then we could have a modifier that would be something like:
```swift
Text("Hello, world!")
  ._domTextSanitizer { $0 }
VStack {
  Text("A large View hierarchy")
}
  ._domTextSanitizer { $0 }
```

carson-katri: Also, should this be in `TokamakStaticHTML`?

NativeDemo/TokamakDemo.xcodeproj/project.pbxproj
```diff
@@ -450,6 +453,7 @@
 			buildActionMask = 2147483647;
 			files = (
 				D1E5FDAD24C1D57000E7485E /* TokamakShim.swift in Sources */,
+				63F2511826BC0CD400CAE212 /* TokamakDOMAdditions.swift in Sources */,
```
MaxDesiatov, line 456: Do we need this in the native demo project? I imagine it's completely unused there, right?

ezraberch: The way I've implemented this is by adding new constructors to `Text`, located in `TokamakStaticHTML`. The Demo uses this new functionality. However, since non-HTML environments don't see these new constructors, the Demo will no longer compile.

The solution is to add dummy functions into TokamakShim. That's the purpose of TokamakDOMAdditions.swift. These don't actually sanitize (as it is unnecessary in non-HTML environments); they just return a normal `Text`.

MaxDesiatov: Makes sense, thanks!

MaxDesiatov (dismissed):
Overall this seems legit. I don't have a strong opinion about having `TokamakDOMAdditions.swift` in TokamakShim. I guess as we have that module as the recommended way to import Tokamak, then it should be fine.

ezraberch:
@MaxDesiatov, do you have an opinion on whether it would be better to implement this as a modifier as @carson-katri suggested?

MaxDesiatov:
I'd say it looks nicer since it feels more "additive" than having additional initializers. It seems to me that the main benefit of the modifier approach is that it's more composable, allowing you to stick it on any view, not just `Text`. But I don't know how hard it would be to implement it that way.

Sources/TokamakShim/TokamakDOMAdditions.swift
ezraberch: I've changed it as you suggested. `_domTextSanitizer`will now set an environment variable, which the renderer will use to sanitize.

Sources/TokamakShim/TokamakDOMAdditions.swift
```diff
@@ -0,0 +1,31 @@
...
+#if !os(WASI)
+public enum Sanitizers {
+  public enum HTML {
+    public static let insecure: (String) -> String = { $0 }
+    public static let encode: (String) -> String = { $0 }
+  }
+}
+
+public extension View {
+  func _domTextSanitizer(_ sanitizer: @escaping (String) -> String) -> some View {
```
carson-katri, line 27: I feel like we should just leave this modifier and the sanitizers off of other platforms instead of providing an inert modifier.

We already have `_dom*` modifiers and they only exist in TokamakDOM. SwiftUI also has modifiers only available on macOS or only on iOS, and they expect you to wrap them with `#if os(_)` (which is made easier in (I believe) 5.5? since they can be placed around a single modifier).

MaxDesiatov (approved):
👍

PR #444, opened 2021-09-01
Add HTML renderer support for AngularGradient

### LINKED ISSUE #443: UnitPoint constants don't match SwiftUI
Some of the static constants in `UnitPoint` do not match the values in SwiftUI. Basically, the y values are flipped (0->1 and 1->0). `.zero` is an exception. There is code elsewhere which flips the y-values back before rendering, but this has the effect of making `UnitPoint`s which do not use these constants (plus `.zero`) render incorrectly.

For example, consider the following
```swift
Rectangle()
  .fill(
    RadialGradient(
      colors: [.red, .green, .blue],
      center: .zero,
      startRadius: 50,
      endRadius: 100
    )
  )
  .frame(width: 300, height: 100)
```

In SwiftUI, the red is in the upper left. In Tokamak, it is in the lower left. A similar difference in behavior can be seen if you use `UnitPoint(x: 1.0, y: 1.0)` as the `center`.

Is there a reason `UnitPoint` was implemented this way?

ezraberch:
This PR adds HTML renderer support for `AngularGradient`. Since SVG gradients do not support angular gradients, CSS `conic-gradient` is used instead.

Additionally, UnitPoint constants have been changed to match SwiftUI (fixes #443).


MaxDesiatov:
Do snapshot tests pass for you locally? There's a mismatch, so I wonder if this is reproducible for you, or maybe we should tweak flags passed to Edge again? Some mismatches are caused by bitness and/or resolution inconsistencies on different devices, so I had to hardcode those paramaters, but maybe there's still something missing?

ezraberch:
They do pass locally, on both 12.5.1 and 13 beta. It seems the upload failed screenshots step isn't working, so I can't see what exactly the difference is.

MaxDesiatov:
I think we need to change the pattern here from `RenderingTests` to just `'*Tests'` (with single-quotes to avoid the in-place glob expansion) https://github.com/TokamakUI/Tokamak/blob/e1fcd180d7fac3632c8cfd8511149fe52f49d5ae/.github/workflows/ci.yml#L40

This line also needs to be updated, I hope this one supports glob star expansions too https://github.com/TokamakUI/Tokamak/blob/e1fcd180d7fac3632c8cfd8511149fe52f49d5ae/.github/workflows/ci.yml#L64

.github/workflows/ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           # avoid building unrelated products for testing by specifying the test product explicitly
           swift build --product TokamakPackageTests
           `xcrun --find xctest` .build/debug/TokamakPackageTests.xctest ||
-            (cp -r `find /var/folders -iname RenderingTests` . ; exit 1)
+            (cp -r /var/folders/*/*/*/*Tests . ; exit 1)
```
MaxDesiatov, line 40: Double star should be able to traverse all nested directories and is hopefully more reliable.

```suggestion
            (cp -r /var/folders/**/*Tests . ; exit 1)
```

ezraberch: That didn't work. Going back to the single wildcard does work, though these is a bit extra in there. Images look identical to me.

ezraberch:
I've updated the test to allow either version of the image. Version A is what I get locally. Version B is what the CI gets. They are indistinguishable visually.

MaxDesiatov:
Did you try to use [`compare`](https://imagemagick.org/script/compare.php) (available if you do `brew install imagemagick`) to generate a pixel diff for these images? Maybe that could point to what exactly is going on there?

ezraberch:
I did; the one place it shows a difference isn't even in the new angular gradient section.

result

MaxDesiatov (approved):
LGTM 👍

Issue #445, opened 2021-09-03
NavigationView causes Error: Detected stack-buffer-overflow


alobaili:
**Describe the bug**
I'm just starting exploring Tokamak and created a project using the tokamak template as described in the readme.

I added a simple `NavigationView` to the body of `ContentView` as follows:
```swift
import TokamakDOM

struct TokamakApp: App {
    var body: some Scene {
        WindowGroup("Tokamak App") {
            ContentView()
        }
    }
}

struct ContentView: View {
    var body: some View {
        NavigationView {
            Text("Hello")
                .foregroundColor(Color.red)
        }
    }
}

// @main attribute is not supported in SwiftPM apps.
// See https://bugs.swift.org/browse/SR-12683 for more details.
TokamakApp.main()
```

Saving the file to trigger a reload shows a blank white screen. The Google Chrome console prints the following error:
```
Error: Detected stack-buffer-overflow.
    at report_stack_overflow (dev.js:85)
    at <anonymous>:wasm-function[86248]:0x180c3d1
    at (anonymous namespace)::NodePrinter::print(swift::Demangle::__runtime::Node*, bool) (<anonymous>:wasm-function[43960]:0x8e7d48)
    at (anonymous namespace)::NodePrinter::printEntity(swift::Demangle::__runtime::Node*, bool, (anonymous namespace)::NodePrinter::TypePrinting, bool, __swift::__runtime::llvm::StringRef, int, __swift::__runtime::llvm::StringRef) (<anonymous>:wasm-function[43962]:0x8edaf5)
    at (anonymous namespace)::NodePrinter::print(swift::Demangle::__runtime::Node*, bool) (<anonymous>:wasm-function[43960]:0x8e8e09)
    at (anonymous namespace)::NodePrinter::print(swift::Demangle::__runtime::Node*, bool) (<anonymous>:wasm-function[43960]:0x8e8d48)
    at (anonymous namespace)::NodePrinter::print(swift::Demangle::__runtime::Node*, bool) (<anonymous>:wasm-function[43960]:0x8eb521)
    at (anonymous namespace)::NodePrinter::print(swift::Demangle::__runtime::Node*, bool) (<anonymous>:wasm-function[43960]:0x8e8d48)
    at (anonymous namespace)::NodePrinter::print(swift::Demangle::__runtime::Node*, bool) (<anonymous>:wasm-function[43960]:0x8e920f)
[...]

PR #446, opened 2021-09-10
Refactor NavigationView

ezraberch:
This PR refactors `NavigationView` in the `TokamakStaticHTML` renderer. While this should have no functional effect, it fixes the crash in #445. This is likely due to an underlying bug in [swiftwasm](https://github.com/swiftwasm/swift).

### DIFF
--- a/Sources/TokamakStaticHTML/Views/Navigation/NavigationView.swift
+++ b/Sources/TokamakStaticHTML/Views/Navigation/NavigationView.swift
@@ -15,6 +15,39 @@
 import TokamakCore
 
 extension NavigationView: _HTMLPrimitive {
+  private struct NavBody: View {
+    let toolbarContent: [AnyToolbarItem]
+    let title: AnyView?
+    let view: NavigationView
+
+    init(_ toolbarContent: [AnyToolbarItem], _ title: AnyView?, _ view: NavigationView) {
+      self.toolbarContent = toolbarContent
+      self.title = title
+      self.view = view
+    }
+
+    var body: some View {
+      HTML("div", ["class": "_tokamak-toolbar-content _tokamak-toolbar-leading"]) {
+        view.items(from: toolbarContent, at: .navigationBarLeading)
+        view.items(from: toolbarContent, at: .navigation)
+        title
+          .font(.headline)
+        view.items(from: toolbarContent, at: .navigationBarTrailing)
+        view.items(from: toolbarContent, at: .automatic, .primaryAction)
+        view.items(from: toolbarContent, at: .destructiveAction)
+          .foregroundColor(.red)
+      }
+      HTML("div", ["class": "_tokamak-toolbar-content _tokamak-toolbar-center"]) {
+        view.items(from: toolbarContent, at: .principal, .status)
+      }
+      HTML("div", ["class": "_tokamak-toolbar-content _tokamak-toolbar-trailing"]) {
+        view.items(from: toolbarContent, at: .cancellationAction)
+        view.items(from: toolbarContent, at: .confirmationAction)
+          .foregroundColor(.accentColor)
+      }
+    }
+  }
+
   @_spi(TokamakStaticHTML)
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
Just a quick note: if the workaround in https://github.com/TokamakUI/Tokamak/issues/367#issuecomment-829229300 also makes the crash go away (I didn't have time to verify that myself yet), this isn't a bug in SwiftWasm, but the fact that Tokamak consumes a lot of stack memory, which causes a stack overflow.

jhoughjr:
I was able to get around that early on by passing some stack size arg in but don’t recall what it was.
I noted that doubling it again when it happened did not work at all.

ezraberch:
> Just a quick note: if the workaround in [#367 (comment)](https://github.com/TokamakUI/Tokamak/issues/367#issuecomment-829229300) also makes the crash go away (I didn't have time to verify that myself yet), this isn't a bug in SwiftWasm, but the fact that Tokamak consumes a lot of stack memory, which causes a stack overflow.

That does make it work, but here's why I think there's still an underlying wasm issue.

If the problematic code is moved to an unreachable code path...

```swift
              if toolbarContent.isEmpty {
                HTML("div", ["class": "_tokamak-toolbar-content _tokamak-toolbar-leading"]) {
                  title.font(.headline)
                }
              } else if true {
                HTML("div", ["class": "_tokamak-toolbar-content _tokamak-toolbar-leading"]) {
                  title.font(.headline)
                }
              } else {
                HTML("div", ["class": "_tokamak-toolbar-content _tokamak-toolbar-leading"]) {
                  items(from: toolbarContent, at: .navigationBarLeading)
                  items(from: toolbarContent, at: .navigation)
                  title
                    .font(.headline)
                  items(from: toolbarContent, at: .navigationBarTrailing)
                  items(from: toolbarContent, at: .automatic, .primaryAction)
                  items(from: toolbarContent, at: .destructiveAction)
                    .foregroundColor(.red)
                }
                HTML("div", ["class": "_tokamak-toolbar-content _tokamak-toolbar-center"]) {
                  items(from: toolbarContent, at: .principal, .status)
                }
                HTML("div", ["class": "_tokamak-toolbar-content _tokamak-toolbar-trailing"]) {
                  items(from: toolbarContent, at: .cancellationAction)
                  items(from: toolbarContent, at: .confirmationAction)
                    .foregroundColor(.accentColor)
                }
              }
```
`NavigationView.swift:39:17: warning: will never be executed`

...then the stack overflow remains. However, commenting out the code in the unreachable block will eliminate the overflow.

MaxDesiatov:
My hunch that's how result builders work. `if` statement there is transformed into an expression, which allocates all of the views in both branches to yield a value of `_ConditionalContent`. In that sense there's no truly "unreachable" blocks with result builders, the only way to prevent them from being eagerly evaluated and blowing up the stack is to hide them in a separate `View` type, or to comment them out completely.

Creating a new `View` type makes that view "lazy", meaning that its `body` content will only evaluate when that view is supposed to be rendered.

ezraberch:
To test your theory, I used the following function:

```swift
  func printAndReturnText(_ text: String) -> Text {
    print(text)
    return Text(text)
  }
```

Then I added a call in each of the three code paths. With TokamakDemo, I get just the second code path. With the minimal test case, I get no print statements at all.

MaxDesiatov:
Hm, fair enough. @kateinoigakukun do you have any ideas? Could this be something related to how function builders work, or is there something else going on that isn't a simple stack overflow?

MaxDesiatov (approved):
On a second thought, merging this as a temporary solution is definitely worth it 👍

kateinoigakukun:
Sorry for slow response. I'll take a look later but it's ok to merge for now.

revolter:
> it fixes the crash in #445

For me, this PR didn't actually fix the crash in #445.

revolter:
For me, this PR didn't actually fix #445.

PR #449, opened 2021-09-19
Add Canvas and TimelineView to DOM renderer

carson-katri:
This is built on the [HTML canvas](https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API), and is compatible with the iOS 15 [`Canvas`](https://developer.apple.com/documentation/swiftui/canvas) view. `TimelineView` with an `.animation` schedule only runs with `Canvas` at the moment with a special implementation using `requestAnimationFrame`, but could be adjusted to work with other views as well.

Here's a demo of confetti using `Canvas`:

https://user-images.githubusercontent.com/13581484/133929292-4b98926c-3e0a-468c-ad23-28d8a87e3c77.mov

### DIFF
--- a/Sources/TokamakCore/Shapes/Ellipse.swift
+++ b/Sources/TokamakCore/Shapes/Ellipse.swift
@@ -27,7 +27,19 @@ public struct Ellipse: Shape {
 
 public struct Circle: Shape {
   public func path(in rect: CGRect) -> Path {
-    .init(storage: .ellipse(rect), sizing: .flexible)
+    .init(
+      storage: .ellipse(
+        .init(
+          x: rect.origin
+            .x + (rect.width > rect.height ? (rect.width / 2) - (rect.width - rect.height) : 0),
+          y: rect.origin
+            .y + (rect.height > rect.width ? (rect.height / 2) - (rect.height - rect.width) : 0),
+          width: min(rect.width, rect.height),
+          height: min(rect.width, rect.height)
+        )
+      ),
+      sizing: .flexible
+    )
   }
 
   public init() {}
--- a/Sources/TokamakCore/Shapes/Ellipse.swift
+++ b/Sources/TokamakCore/Shapes/Ellipse.swift
@@ -30,10 +30,9 @@ public struct Circle: Shape {
     .init(
       storage: .ellipse(
         .init(
-          x: rect.origin
-            .x + (rect.width > rect.height ? (rect.width / 2) - (rect.width - rect.height) : 0),
-          y: rect.origin
-            .y + (rect.height > rect.width ? (rect.height / 2) - (rect.height - rect.width) : 0),
+          // Center the circle in the rect.
+          x: rect.origin.x + (rect.width > rect.height ? (rect.width - rect.height) / 2 : 0),
+          y: rect.origin.y + (rect.height > rect.width ? (rect.height - rect.width) / 2 : 0),
           width: min(rect.width, rect.height),
           height: min(rect.width, rect.height)
         )
--- a/Sources/TokamakCore/ViewTraits/TagValueTraitKey.swift
+++ b/Sources/TokamakCore/ViewTraits/TagValueTraitKey.swift
@@ -1,0 +1,47 @@
+// Copyright 2020 Tokamak contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//  Created by Carson Katri on 9/23/21.
+//
+
+import Foundation
+
+public extension View {
+  @inlinable
+  func tag<V>(_ tag: V) -> some View where V: Hashable {
+    _trait(TagValueTraitKey<V>.self, .tagged(tag))
+  }
+
+  @inlinable
+  func _untagged() -> some View {
+    _trait(IsAuxiliaryContentTraitKey.self, true)
+  }
+}
+
+@usableFromInline
+struct TagValueTraitKey<V>: _ViewTraitKey where V: Hashable {
+  @usableFromInline
+  enum Value {
+    case untagged
+    case tagged(V)
+  }
+
+  @inlinable static var defaultValue: Value { .untagged }
+}
+
+@usableFromInline
+struct IsAuxiliaryContentTraitKey: _ViewTraitKey {
+  @inlinable static var defaultValue: Bool { false }
+  @usableFromInline typealias Value = Bool
+}
--- a/Sources/TokamakCore/ViewTraits/TagValueTraitKey.swift
+++ b/Sources/TokamakCore/ViewTraits/TagValueTraitKey.swift
@@ -1,4 +1,4 @@
-// Copyright 2020 Tokamak contributors
+// Copyright 2021 Tokamak contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/Sources/TokamakCore/Views/Canvas/TimelineView.swift
+++ b/Sources/TokamakCore/Views/Canvas/TimelineView.swift
@@ -166,7 +166,7 @@ public extension TimelineSchedule where Self == AnimationTimelineSchedule {
 
 public struct AnimationTimelineSchedule: TimelineSchedule {
   private let minimumInterval: Double?
-  private let paused: Bool
+  public let _paused: Bool
 
   public struct Entries: Sequence, IteratorProtocol {
     var date: Date
@@ -185,10 +185,10 @@ public struct AnimationTimelineSchedule: TimelineSchedule {
 
   public init(minimumInterval: Double? = nil, paused: Bool = false) {
     self.minimumInterval = minimumInterval
-    self.paused = paused
+    _paused = paused
   }
 
   public func entries(from startDate: Date, mode: TimelineScheduleMode) -> Entries {
-    Entries(date: startDate, minimumInterval: minimumInterval, paused: paused)
+    Entries(date: startDate, minimumInterval: minimumInterval, paused: _paused)
   }
 }
--- a/Sources/TokamakCore/Views/Canvas/TimelineView.swift
+++ b/Sources/TokamakCore/Views/Canvas/TimelineView.swift
@@ -1,0 +1,196 @@
+// Copyright 2020-2021 Tokamak contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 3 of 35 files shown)


Sources/TokamakCore/Shapes/Ellipse.swift
```diff
@@ -27,7 +27,19 @@ public struct Ellipse: Shape {
 
 public struct Circle: Shape {
   public func path(in rect: CGRect) -> Path {
-    .init(storage: .ellipse(rect), sizing: .flexible)
+    .init(
+      storage: .ellipse(
+        .init(
+          x: rect.origin
+            .x + (rect.width > rect.height ? (rect.width / 2) - (rect.width - rect.height) : 0),
```
MaxDesiatov, line 34: Can you add some comments in code explaining why this change is needed?

Sources/TokamakDOM/Views/Canvas/Canvas.swift
```diff
@@ -0,0 +1,514 @@
...
+        "width": "\(proxy.size.width * CGFloat(devicePixelRatio))",
+        "height": "\(proxy.size.height * CGFloat(devicePixelRatio))",
+      ])
+        ._domRef($canvas)
+        .onAppear(perform: draw)
+        ._onUpdate(perform: draw)
+    }
+  }
+
+  private func draw() {
```
MaxDesiatov, line 42: Would there be an easy way to split this into multiple functions to make it more readable?

Sources/TokamakDOM/Views/Canvas/Canvas.swift
```diff
@@ -0,0 +1,514 @@
...
+    case .luminanceToAlpha:
+      break
+    case let .blur(radius, _):
+      canvasContext.filter = .string("\(previousFilter) blur(\(radius)px)")
+    case .alphaThreshold:
+      break
+    }
+  }
+
+  private func pushPath(_ path: Path, in canvasContext: JSObject) {
```
MaxDesiatov, line 212: ditto

Sources/TokamakDOM/Views/Canvas/Canvas.swift
```diff
@@ -0,0 +1,514 @@
...
+        }
+      case .kerning, .tracking, .rounded, .baseline, .strikethrough,
+           .underline: break // Not supported in <canvas>.
+      }
+    }
+
+    canvas
+      .font =
+      .string(
+        "\(style) \(variant) \(weight) \(size ?? "17")pt \(lineHeight) \(family ?? Font.Design.default.families.joined(separator: " "))"
```
MaxDesiatov, line 342:
```suggestion
        """
        \(style) \(variant) \(weight) \(size ?? "17")pt \(lineHeight) \(
          family ?? Font.Design.default.families.joined(separator: " ")
        )
        """
```

Sources/TokamakDOM/Views/Canvas/Canvas.swift
```diff
@@ -0,0 +1,514 @@
...
+  ) -> JSValue {
+    switch self {
+    case let .color(color):
+      return .string(AnyColorBox.ResolvedValue(
+        red: color.red,
+        green: color.green,
+        blue: color.blue,
+        opacity: color.opacity * Double(opacity),
+        space: color.space
+      ).cssValue)
```
MaxDesiatov, line 426: Maybe this could moved into some extension on `AnyColorBox.ResolvedValue` that has an initializer taking a `color` value and supplied `opacity`. That would reduce these 6 lines to 1 with just that helper initializer call, also potentially resolving the linter warning.

Sources/TokamakDOM/Views/Canvas/Canvas.swift
```diff
@@ -0,0 +1,514 @@
...
+  }
+}
+
+extension Canvas: DOMPrimitive {
+  public var renderedBody: AnyView {
+    AnyView(_Canvas(parent: self))
+  }
+}
+
+private extension GraphicsContext.Shading {
```
MaxDesiatov, line 353: Extensions below this line could into a separate file? Yes, they would no longer stay `private`, but it would potentially improve readability and overall navigation through the project, preventing files from becoming larger.

MaxDesiatov (commented):
Marvellous! 🎊

Other than a few linter nits, maybe there's a way to add an HTML snapshot test? Albeit primitive it could ensure that at least the `<canvas>` tag is rendered.

MaxDesiatov:
Disregard the snapshot test comment, I realize this won't be possible right now because we can't test `TokamakDOM` directly, and this is implemented in `TokamakDOM`.

carson-katri:
Haven't done much cleanup yet, but did add support for symbols:
```swift
Canvas { context, size in
  // Draw axes on the canvas.
  context.stroke(Rectangle().path(in: .init(x: size.width / 2, y: 0, width: 1, height: size.height)), with: .color(.green))
  context.stroke(Rectangle().path(in: .init(x: 0, y: size.height / 2, width: size.width, height: 1)), with: .color(.green))
  
  // Resolve and draw the symbols, looking them up by tag.
  guard let aSymbol = context.resolveSymbol(id: "a"),
        let bSymbol = context.resolveSymbol(id: "b") else { return }
  context.draw(aSymbol, at: .init(x: size.width / 2, y: size.height / 2), anchor: .topLeading)
  context.draw(bSymbol, at: .init(x: size.width / 2, y: size.height / 2), anchor: .bottomTrailing)
} symbols: {
  Text("A")
    .frame(width: 100, height: 100)
    .background(Color.red)
    .cornerRadius(10)
    .tag("a") // Tagged with "a"
  Text("B")
    .frame(width: 100, height: 100)
    .background(Color.blue)
    .tag("b") // Tagged with "b"
}
```
Screen Shot 2021-09-23 at 2 59 15 PM

To look up Views by their tag, I added support for `_VariadicView` (matching SwiftUI's) which provides access to child views as a `RandomAccessCollection`, as well as access to `_ViewTraitKey`s. The basic usage in Canvas is like so:
```swift
private struct SymbolResolverLayout<ID: Hashable>: _VariadicView.ViewRoot {
  let id: ID

  func body(children: _VariadicView.Children) -> some View {
    ForEach(children) {
      if case let .tagged(tag) = $0[TagValueTraitKey<ID>.self],
         tag == id
      {
        $0
      }
    }
  }
}

// Then used like so:
_VariadicView.Tree(SymbolResolverLayout(id: id)) {
  _storage.symbols
}
```

carson-katri:
Off-topic, but `_VariadicView` could be used to solve some issues with `Picker`. It can allow for the use of `tag(_:)` to specify items alongside `ForEach`:
```swift
Picker("Choose...", selection: $selection) {
  ForEach(items) { Text($0.title) }
  Text("Clear").tag(Item.clear)
}
```
or for re-tagging:
```swift
Picker("Choose...", selection: $value) {
  ForEach(items) {
    Text($0.title)
      .tag($0.value)
  }
}
```

MaxDesiatov:
Would this be ready for review soon? It would be great to keep the PR relatively small, and then additional features could be implemented in a separate PR, for easier review, and to have separate commits for such features in the `main` branch history.

carson-katri:
Yes, it can reviewed now.

Sources/TokamakCore/ViewTraits/TagValueTraitKey.swift
```diff
@@ -0,0 +1,47 @@
+// Copyright 2020 Tokamak contributors
```
MaxDesiatov, line 1:
```suggestion
// Copyright 2021 Tokamak contributors
```

Sources/TokamakDOM/Views/Canvas/ImageOperations.swift
```diff
@@ -0,0 +1,98 @@
...
+    set { values[path] = newValue }
+  }
+}
+
+extension _Canvas {
+  func resolveImage(
+    _ image: Image,
+    _ environment: EnvironmentValues
+  ) -> GraphicsContext.ResolvedImage {
+    // FIXME: We don't have a way of calculating the size, since we need to wait for the image to load.
```
MaxDesiatov, line 36:
```suggestion
    // FIXME: We don't have a way of calculating the size, since we need to wait for the image
    // to load.
```

Sources/TokamakDOM/Views/Canvas/SymbolOperations.swift
```diff
@@ -0,0 +1,115 @@
...
+          Double(rect.size.width), Double(rect.size.height)
+        )
+      case let .at(point, anchor):
+        _ = canvasContext.drawImage!(
+          img,
+          Double(point.x - (anchor.x * symbol.size.width)),
+          Double(point.y - (anchor.y * symbol.size.height))
+        )
+      }
+//      _ = JSObject.global.URL.function!.revokeObjectURL!(objectURL)
```
MaxDesiatov, line 90: Can this be cleaned up?

Sources/TokamakDOM/Views/Canvas/TextOperations.swift
```diff
@@ -0,0 +1,137 @@
...
+        }
+      case .kerning, .tracking, .rounded, .baseline, .strikethrough,
+           .underline: break // Not supported in <canvas>.
+      }
+    }
+
+    canvas
+      .font =
+      .string(
+        "\(style) \(variant) \(weight) \(size ?? "17")pt \(lineHeight) \(family ?? Font.Design.default.families.joined(separator: " "))"
```
MaxDesiatov, line 134:
```suggestion
        """
        \(style) \(variant) \(weight) \(size ?? "17")pt \(lineHeight) \(
          family ?? Font.Design.default.families.joined(separator: " ")
        )
        """
```

Sources/TokamakDemo/CanvasDemo.swift
```diff
@@ -0,0 +1,96 @@
+// Copyright 2019-2020 Tokamak contributors
```
MaxDesiatov, line 1:
```suggestion
// Copyright 2021 Tokamak contributors
```

MaxDesiatov (commented):
Looks great! Just a few nits

Sources/TokamakCore/Views/Canvas/TimelineView.swift
```diff
@@ -0,0 +1,196 @@
...
+
+  public struct Entries: Sequence, IteratorProtocol {
+    internal let startDate: Date
+    internal let minimumInterval: Double?
+    internal let paused: Bool
+
+    public mutating func next() -> Date? {
+      guard !paused else { return nil }
+      guard let minimumInterval = minimumInterval else {
+        return .init()
```
ezraberch, line 188: According to the spec, "Pass nil to let the system pick an appropriate update interval." Instead, it looks like this is just updating as quickly as possible when `minimumInterval` is nil. This isn't good for performance.

Additionally, `next()` is a mutating function. It's supposed to change the struct to return a different value next time it's called.

Sources/TokamakDemo/CanvasDemo.swift
```diff
@@ -0,0 +1,96 @@
...
+
+  static let shape = Rectangle()
+
+  var body: some View {
+    TimelineView(AnimationTimelineSchedule.animation) { timeline in
+      Canvas { context, size in
+        let elapsed = CGFloat(
+          timeline.date.timeIntervalSince1970 - startDate
+            .timeIntervalSince1970
+        ) * 10
```
ezraberch, line 69: It would be better if the demo looped rather than going to an empty screen after a few seconds.

```suggestion
        ).truncatingRemainder(dividingBy: 4.0) * 10
```

ezraberch (approved):
Looks good. 👍

PR #461, opened 2021-12-31
Fix `rootEnvironment` not merged with `.defaultEnvironment`

regexident:
While spelunking through the implementation of `EnvironmentValues` and its uses in Tokamak I noticed that a couple of functions were accepting a `rootEnvironment: EnvironmentValues? = nil` argument but then discarding whatever value you pass to it in favor of `.defaultEnvironment`.

This seems wrong to me? 🤔

### DIFF
--- a/Sources/TokamakCore/Environment/EnvironmentValues.swift
+++ b/Sources/TokamakCore/Environment/EnvironmentValues.swift
@@ -44,6 +44,7 @@ public struct EnvironmentValues: CustomStringConvertible {
     }
   }
 
+  @_spi(TokamakCore)
   public mutating func merge(_ other: Self?) {
     if let other = other {
       values.merge(other.values) { existing, new in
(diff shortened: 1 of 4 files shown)


carson-katri:
Good catch. I think you may still need to add in the values `.defaultEnvironment` sets up if they are not already setup by the `rootEnvironment` though, specifically `_ToggleStyleKey`, `.colorScheme`, `._defaultAppStorage`, and `_DefaultSceneStorageProvider`.

regexident:
Let me know if you'd rather have the merging logic as a private implementation detail. I figured since effectively every backend/renderer implementation would need it it might be worth providing a canonical method for it. Providing both `.merge(_:)` and `.merging(_:)` might a bit overkill though.

carson-katri:
Since this isn't part of SwiftUI, may be best to make it internal with `@_spi` so the renderers can access it but users don't.

regexident:
Absolutely! Fixed.

carson-katri (changes requested):
I think you still need to add `@_spi(TokamakCore)` to some imports to fix the builds. Once that's fixed this LGTM!

regexident:
Fixed!

MaxDesiatov (approved):
LGTM, thanks!

carson-katri (approved):

Issue #441, opened 2021-08-18
(H/V)Stack doesn't align elements correctly if used with a Spacer


Joannis:
**Describe the bug**
When using a (V/H)Stack with a spacer inbetween two elements, the elements should take up their natural size. Instead, they're taking up an equal third of the space, and are aligned at their natural 'start'. Which is top for VStack and left for HStack.

**To Reproduce**
```swift
struct ContentView: View {
    var body: some View {
        HStack {
            Text("HI")
            Spacer()
            Text("HI")
        }
    }
}
```

**Expected behavior**
The second "HI" should be aligned to the far right.

**Screenshots**
<kbd>
<img width="708" src="https://user-images.githubusercontent.com/1951674/129876798-607f8235-b0f6-4c61-826a-1040641b3507.png">
</kbd>

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Safari
 - Version of the browser  15.0 (17612.1.24.1.4)
 - Version of Tokamak [e.g. 0.6.1]

carson-katri:
This is probably because we started using grid for stacks instead of flex box because of the `gap` property. However, I think flexbox supports the gap property in most browsers, so could we maybe switch back?

Joannis:
Correct, it's due to grids. But I'm not sure what a good alternative is.

carson-katri:
Here's is the browser support for Flexbox gap for reference: https://caniuse.com/flexbox-gap

SwiftCoderJoe:
`flexbox-gap` looks good enough to me. 

The only major browsers based on relative userbase size that I could see being a problem are the older Safari browsers (Both desktop and iOS), but considering Apple platforms have a history of having fast adoption of newer OS versions, I wouldn't be too worried about those user-groups staying around long. 

I'd vote to switch back to `flex` considering how high the standard adoption is and how often aligning things using `spacer()` can be useful.

PR #466, opened 2022-02-05
Change HStack and VStack to Flex

SwiftCoderJoe:
Flex fixes a lot of issues with Spacer (#441) and alignment. I tested alignment with the demo and made a change to the spacer demo to show centering using spacers.

It's worth noting that the minimum requirements for `gap` on `flexbox` is lower than the minimum requirements for Tokamak already, so there shouldn't be any compatibility issues. (https://caniuse.com/flexbox-gap)

I also added `box-sizing: border-box` to the default Tokamak style which fixed some issues in some demos (Namely the Spacer demo, which is why it's included in this PR)

This is my first contribution to Tokamak so I'd appreciate review and feedback!

### DIFF
--- a/Sources/TokamakDemo/Layout/SpacerDemo.swift
+++ b/Sources/TokamakDemo/Layout/SpacerDemo.swift
@@ -23,6 +23,8 @@ struct SpacerDemo: View {
       HStack {
         Text("Left side.")
         Spacer()
+        Text("Center.")
+        Spacer()
         Text("Right side.")
       }
       Spacer()
--- a/Sources/TokamakDemo/Layout/StackDemo.swift
+++ b/Sources/TokamakDemo/Layout/StackDemo.swift
@@ -18,14 +18,17 @@ import TokamakShim
 struct StackDemo: View {
   @State private var horizontalSpacing: CGFloat = 8
   @State private var verticalSpacing: CGFloat = 8
+  @State private var horizontalAlignment: Int = 0
 
   var body: some View {
     VStack(spacing: verticalSpacing) {
+
       Text("Horizontal Spacing")
       Slider(value: $horizontalSpacing, in: 0...100)
 
       Text("Vertical Spacing")
       Slider(value: $verticalSpacing, in: 0...100)
+
       HStack(spacing: horizontalSpacing) {
         Rectangle()
           .fill(Color.red)
@@ -35,6 +38,10 @@ struct StackDemo: View {
           .fill(Color.green)
           .frame(width: 100, height: 100)
       }
+
+      Rectangle()
+          .fill(Color.yellow)
+          .frame(width: 100, height: 100)
     }
   }
 }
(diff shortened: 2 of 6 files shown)


MaxDesiatov (changes requested):
Overall this is great, and I don't want to discourage you from continued work on this. Efforts to clean up the layout code are greatly appreciated!

There are two test failures though, you can go to [the build summary page](https://github.com/TokamakUI/Tokamak/actions/runs/1800641428) and download "Failed snapshots" archive at the bottom.

I don't think `testGradients-A.1.png` has any meaningful changes, so it's probably fine to delete the old version and commit the new one. Maybe @carson-katri and @ezraberch as original implementors of gradients can see a substantial difference and point to what's causing it.

Here's the original:
testGradients-A 1
and the new version:
testGradients-A 1

`testMaterial.1.png` change is obvious though, compare the original
testMaterial 1
to the new version
testMaterial 1

I'm not 100% sure what's causing it, but it would be great to keep this test passing.

Also, it's highly desirable to add more snapshot tests to cover the new behavior and to prevent from breaking in the future. I guess new tests could be based on some existing demo code, but ideally shouldn't contain any text, since font rendering can change in unexpected ways on different platforms.

ezraberch:
I'm not seeing a failure in the Gradients test, either on the CI or locally. From the CI report:
```
Test Case '-[TokamakStaticHTMLTests.VisualRenderingTests testGradients]' started.
[921](https://github.com/TokamakUI/Tokamak/runs/5080472683?check_suite_focus=true#step:3:921)
[0205/235103.645806:INFO:headless_shell.cc(653)] Written to file screenshot.png.
[922](https://github.com/TokamakUI/Tokamak/runs/5080472683?check_suite_focus=true#step:3:922)
[0205/235104.949274:INFO:headless_shell.cc(653)] Written to file screenshot.png.
[923](https://github.com/TokamakUI/Tokamak/runs/5080472683?check_suite_focus=true#step:3:923)
Test Case '-[TokamakStaticHTMLTests.VisualRenderingTests testGradients]' passed (1.826 seconds).
```

If you're just going by the presence of the image in the failed screenshots upload, that's a side effect of that test passing if the image matches either of two references images (see #444).

MaxDesiatov:
Oh, that's great, thanks for checking! Then it's just one visual snapshot, and probably also a couple of HTML snapshots. And the latter can just have new versions committed.

SwiftCoderJoe:
Sorry about that! All the tests passed on my local machine, which is a Linux environment. As far as I can tell, these tests are MacOS only. It might take me a while to set up my MacOS environment, so please be patient while I fix this.

MaxDesiatov:
No problem at all! These tests are not inherently specific to macOS. Visual tests have Edge browser as a dependency, so in theory they should be runnable on Linux. HTML snapshots don't need any browser at all, so running those on Linux should be even easier.

SwiftCoderJoe:
Hmm, maybe I'm confused about how to run them then. Are they not just through `carton test`?

MaxDesiatov:
These tests don't run in the browser, because snapshot tests aren't supported in the browser due to the lack of filesystem access. That's why we can only test use snapshots for `TokamakStaticHTML` target, which is buildable on host platforms. I wonder if simple `swift test` in the root directory of the repo works for you on Linux?

SwiftCoderJoe:
Currently incredibly busy with school stuff. I'll try picking this back up when work dies down.

PR #471, opened 2022-02-21
Experimental "React Fiber"-like Reconciler

carson-katri:
This PR adds a new reconciler modeled after React Fiber. It has a few *potential* benefits over the current stack reconciler:

1. Non-recursive. This *may* help mitigate some of the stack overflow issues.
2. Reduces use of `AnyView`. We can instead use a `ViewVisitor` to traverse the View tree without type-erasing all of the children.
3. Mounted element updates are batched, which *may* reduce some of the cost of bridging to JavaScript by deferring it until all mutations have been gathered, although I'm not sure if this is the case or not.
4. Can be used to experiment with a custom layout engine that matches SwiftUI 1-1. I've been experimenting with this in the `fiber/layout` branch, but it does not work all that well yet.

We will need to test it to figure out if these are actually of any benefit.

This exists alongside the current `StackReconciler` class and `Renderer` protocol as `FiberReconciler` and `FiberRenderer`. I'd like to get some feedback on this to know if its something that would be worth the effort, and whether or not I've overlooked anything major in the design.

I have only tested it with very simple Views. `App` and `Scene` are not supported, as well as preferences. However, preferences may be simple to implement with this model as we traverse down then back up the tree in the main reconciler loop.

> Note: To try the DOM renderer you must pass the custom `index.html` to Carton, as all DOM mutations are performed on the JavaScript side.


ezraberch:
> * Non-recursive. This _may_ help mitigate some of the stack overflow issues.

To test this, I used the following:

```swift
struct RecursiveView: View {
  let count: Int

  init(_ count: Int) {
    self.count = count
  }

  var body: some View {
    if count == 0 {
      Text("RecursiveView")
    } else {
      RecursiveView(count-1)
    }
  }
}
```

Results after adding it to your demo:

Current reconciler: 400 works, 500 overflows.
New reconciler: 900 works, 1000 loads but button causes overflow, 1100 overflows.

carson-katri:
@ezraberch That's cool, thanks for testing it out! Is that with or without the `stack-size` flag in Package.swift?

ezraberch:
With. If I remove that flag, the breaking point changes to around 30 for current and 700 for new.

MaxDesiatov:
Would fixing 3 remaining tests require any significant changes to this PR?

Sources/TokamakCore/Fiber/Element.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  File.swift
+//
```
MaxDesiatov, line 3: Ditto for the rest of new files.

```suggestion
```

Sources/TokamakCore/Fiber/Element.swift
```diff
@@ -0,0 +1,21 @@
...
+//
+//  Created by Carson Katri on 2/15/22.
+//
+
+/// A reference type that points to a `Renderer`-specific element that has been mounted.
+/// For instance, a DOM node in the `DOMFiberRenderer`.
+public protocol Element: AnyObject {
```
MaxDesiatov, line 10: Do I understand correctly that this roughly corresponds to the `Target` protocol used in the stack reconciler? Either way, I find the `Element` name by itself too generic for keeping it in the top-level namespace of the module. Maybe something like `FiberElement` could work?

Sources/TokamakCore/Fiber/Element.swift
```diff
@@ -0,0 +1,21 @@
...
+  associatedtype Data: ElementData
```
MaxDesiatov, line 11: Similar concern about `Data` name, maybe `Content` could work better? Besides, there's an established precedent for using `Content` name in SwiftUI as a generic parameter.

Sources/TokamakCore/Fiber/Element.swift
carson-katri: Yes, it does correspond to `Target`. `FiberElement` sounds good.

Sources/TokamakCore/Fiber/Fiber.swift
```diff
@@ -0,0 +1,240 @@
...
+    /// The mounted element, if this is a Renderer primitive.
+    var element: Renderer.ElementType?
+    /// The first child node.
+    var child: Fiber?
+    /// This node's right sibling.
+    var sibling: Fiber?
+    /// An unowned reference to the parent node.
+    unowned var parent: Fiber?
+    /// The nearest parent that can be mounted on.
+    unowned var elementParent: Fiber?
```
MaxDesiatov, line 44: Any specific reason this is `unowned` instead of `weak`? I think using the former could be fine, as long as the nuance is explained here in code comments.

Sources/TokamakCore/Fiber/Fiber.swift
```diff
@@ -0,0 +1,240 @@
...
+    /// The cached type information for the underlying `View`.
+    var typeInfo: TypeInfo?
+    /// Boxes that store `State` data.
+    var state: [PropertyInfo: MutableStorage]!
```
MaxDesiatov, line 48: Do we need to distinguish between `nil` and an empty dictionary here? If not, could the implicitly unwrapped optional be avoided by initializing this to an empty dictionary instead?

And overall, I see some of the new properties are IOU, but some are simple optionals. What's the reason for making this distinction?

Sources/TokamakCore/Reflection/Models/PropertyInfo.swift
```diff
@@ -20,7 +20,20 @@
...
+      .offset && lhs.ownerType == rhs.ownerType
+  }
+
+  public func hash(into hasher: inout Hasher) {
+    hasher.combine(name)
+    hasher.combine(ObjectIdentifier(type))
+    hasher.combine(isVar)
+    hasher.combine(offset)
+    hasher.combine(ObjectIdentifier(ownerType))
+  }
```
MaxDesiatov, line 35: Aren't these implementations generated automatically by the compiler? If not, please add a comment above the functions to clarify that.

Sources/TokamakDOM/index.html
```diff
@@ -0,0 +1,82 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+    <title>Tokamak App</title>
+    <script>
+      window.Tokamak = {
```
MaxDesiatov, line 8: Is there any benefit to having this snippet written in JS instead of Swift? Performance benefits for example?

MaxDesiatov (commented):
I love how well documented this is, amazing work! 👏

Just a couple of nits in terms of naming, code comments, and file headers.

Also, did you try adding benchmarks to `TokamakStaticHTMLBenchmark` to compare how well it reconciles when compared to the existing stack reconciler?

carson-katri:
I added some benchmarks to compare the two reconcilers. Here are the results I got running it.

|name                                         |time            |std        |iterations
|---------------------------------------------|----------------|-----------|----------
|render Text (StackReconciler)                |   62699.000 ns |±  35.86 % |     20785
|render Text (FiberReconciler)                |   77655.000 ns |±  38.10 % |     16989
|render ForEach (StackReconciler)             |15578671.000 ns |±   7.64 % |        87
|render ForEach (FiberReconciler)             | 4016959.000 ns |±   7.70 % |       335
|render RecursiveView(1000) (StackReconciler) |44926880.000 ns |±   1.82 % |        31
|render RecursiveView(1000) (FiberReconciler) |20088020.000 ns |±   4.34 % |        69

Also calculated the differences:
|name                                         |Δ               |
|---------------------------------------------|----------------|
|render Text                                  |    +14956.000 ns|
|render ForEach                               | -11561712.000 ns|
|render RecursiveView(100)                    | -24838860.000 ns|

MaxDesiatov:
Splendid! I think a slight regression in `render Text` is acceptable, especially if it improves correctness. If not, it could be fixed in a separate PR in the future. Then it's just a matter of fixing CI, addressing comments, and I'm would be super happy to approve it 🙂

MaxDesiatov:
I've fixed build issues, but I'm not seeing anything on the screen when running `TokamakDemo`, I hope didn't break anything...

Sources/TokamakCore/Fiber/Fiber.swift
carson-katri: I added a doc comment explaining `unowned` on the `parent` property. If you think my reasoning behind it is flawed though, let me know and I'll switch it to `weak`. I'm always a bit unsure of myself on ARC stuff 😅

MaxDesiatov:
Thanks for the updates! There are still a few unresolved comments, and I've requested a review from the team if there are any active team members remaining who could also have a look in the meantime.

Sources/TokamakDOM/index.html
MaxDesiatov: @carson-katri I assume you've addressed this with your benchmarks comment? Could still benefit with a code comment either making a reference to benchmarks or some other explanation.

carson-katri: I actually wasn't able to test that. For now, I've rewritten that in Swift. It definitely simplifies things a bit.

.github/workflows/ci.yml
```diff
@@ -15,14 +15,15 @@ jobs:
...
-    steps:
-      - uses: actions/checkout@v2
-      - uses: swiftwasm/swiftwasm-action@v5.6
-        with:
-          shell-action: carton test
+  #   steps:
+  #     - uses: actions/checkout@v2
+  #     - uses: swiftwasm/swiftwasm-action@v5.6
+  #       with:
+  #         shell-action: carton test
```
MaxDesiatov, line 26: I wonder if this will work now with new `carton test --environment node` introduced in `carton` 0.15? I've updated the GitHub Action to that latest version already.

```suggestion
    steps:
      - uses: actions/checkout@v2
      - uses: swiftwasm/swiftwasm-action@v5.6
        with:
          shell-action: carton test --environment node
```

MaxDesiatov: Yeah, I've verified that `carton test --environment node` works locally on macOS, so I think we should re-enable the tests this way.

MaxDesiatov:
The test failure is most probably a bug in `carton`, I'll get that resolved unless you've already figured this one out 🙂

MaxDesiatov:
Bug is fixed in https://github.com/swiftwasm/carton/pull/350, I'm ready to merge that and tag a new version when approved.

MaxDesiatov (dismissed):
Amazing! 👏

MaxDesiatov:
Demo code still needs to be restored, as far as I understand? I'd prefer that to be done before merging.

MaxDesiatov (approved):

PR #472, opened 2022-02-21
Custom Layout Engine for Fiber Reconciler

carson-katri:
This adds a layout pass to the WIP fiber reconciler, that should in theory be able to replicate SwiftUI layout 1:1 as it’s developed.

Screen Shot 2022-02-21 at 4 13 11 PM

> In this example, `Text` is not measured and is a fixed size of 100x50.

Here are some things that are missing, but should be implemented (perhaps as separate PRs targeting this branch):
- [x] Renderer-specific `Text` measuring
- [ ] `alignmentGuide` modifier
- [ ] Coordinate Spaces
- [x] Disabling layout computations for renderers that don't, or optionally, use them
- [ ] Layouts for all of the existing Views and modifiers, such as `ZStack`, `.background`, `.overlay`, `.frame` etc.


carson-katri:
I've made some revisions, and it now seems to work well with simple examples:

https://user-images.githubusercontent.com/13581484/156600586-894ababe-fe06-42ba-a4a5-c23487a55b75.mov

MaxDesiatov:
Weird to see tests hanging, maybe it's worth restarting them?

carson-katri:
Hm, still hung...

MaxDesiatov:
Do these tests hang for you locally as well?

carson-katri:
Oh, I thought it was one of the tests but its the benchmark that's hanging. It's taking a long time locally as well, so I probably broke something.

carson-katri:
Tests are passing now! 🎉  For reference, here are the results I get from TokamakCore benchmarks with the layout engine enabled vs. disabled:

### Layout Enabled
```
name                             time      std        iterations
----------------------------------------------------------------
mangledName Runtime               0.002 ms ±  24.29 %     836412
typeConstructorName TokamakCore   0.001 ms ±  16.81 %     965430
update wide (StackReconciler)    46.891 ms ±   1.67 %         30
update wide (FiberReconciler)    33.054 ms ±   1.82 %         42
update narrow (StackReconciler)  46.437 ms ±   2.98 %         30
update narrow (FiberReconciler)  33.150 ms ±   3.61 %         42
update deep (StackReconciler)    17.587 ms ±   1.54 %         80
update deep (FiberReconciler)     6.135 ms ±   3.47 %        227
update shallow (StackReconciler)  8.952 ms ±   1.56 %        156
update shallow (FiberReconciler)  3.854 ms ±   3.17 %        363
```

### Layout Disabled
```
name                             time      std        iterations
----------------------------------------------------------------
mangledName Runtime               0.002 ms ±  24.98 %     831827
typeConstructorName TokamakCore   0.001 ms ±  23.00 %     975591
update wide (StackReconciler)    49.103 ms ±   3.19 %         28
update wide (FiberReconciler)    21.304 ms ±   5.15 %         68
update narrow (StackReconciler)  48.522 ms ±   3.33 %         28
update narrow (FiberReconciler)  20.052 ms ±   4.09 %         69
update deep (StackReconciler)    18.120 ms ±   2.90 %         76
update deep (FiberReconciler)     6.254 ms ±   5.24 %        220
update shallow (StackReconciler)  8.950 ms ±   2.11 %        154
update shallow (FiberReconciler)  3.726 ms ±   3.41 %        371
```

carson-katri:
It's not quite ready for review though, still investigating an issue with what I think is `ViewModifiers`. If you have a layout like this, the view inside the modifier (specifically a modifier that has a DOM representation) won't be laid out:
```swift
var body: some View {
  Text("View with modifier (not laid out)")
    .frame(width: 10, height: 10)
  Text("Sibling View without modifier is laid out")
}
```
However, a View body like this (with no sibling) works fine:
```swift
var body: some View {
  Text("View with modifier (is correctly laid out)")
    .frame(width: 10, height: 10)
}
```

Brewfile.lock.json
```diff
@@ -0,0 +1,125 @@
+{
```
MaxDesiatov, line 1: Maybe let's include this file in a separate PR? In that PR I'd also add `swiftwasm/tap/carton` to `Brewfile`.

Sources/TokamakCore/Environment/EnvironmentKey.swift
```diff
@@ -17,11 +17,19 @@ public protocol EnvironmentKey {
   static var defaultValue: Value { get }
 }
 
-protocol EnvironmentModifier {
+public protocol _EnvironmentModifier {
```
MaxDesiatov, line 20: Didn't realize we also have `EnvironmentalModifier`. Should we reconsider the naming of this protocol to avoid any confusion? If there's no other good naming option, we should at least have doc comments that clarify the difference between these protocols.

Sources/TokamakCore/Fiber/AlignmentID.swift
```diff
@@ -0,0 +1,124 @@
+// Copyright 2021 Tokamak contributors
```
MaxDesiatov, line 1: nit
```suggestion
// Copyright 2022 Tokamak contributors
```

Sources/TokamakCore/Fiber/FiberReconciler+TreeReducer.swift
```diff
@@ -0,0 +1,146 @@
+// Copyright 2021 Tokamak contributors
```
MaxDesiatov, line 1: nit
```suggestion
// Copyright 2022 Tokamak contributors
```

Sources/TokamakCore/Fiber/Layout/StackLayoutComputer.swift
```diff
@@ -0,0 +1,148 @@
...
+//
+
+import Foundation
+
+final class StackLayoutComputer: LayoutComputer {
+  let proposedSize: CGSize
+  let axis: Axis
+  let alignment: Alignment
+  let spacing: CGFloat
+  /// A multiplier of `1` for the axis that takes the size of the largest child.
```
MaxDesiatov, line 25: Could you clarify the comment so that it's clear what "a multiplier of `1`" means in this context?

Sources/TokamakCore/Fiber/Layout/StackLayoutComputer.swift
```diff
@@ -0,0 +1,148 @@
+// Copyright 2021 Tokamak contributors
```
MaxDesiatov, line 1:
```suggestion
// Copyright 2022 Tokamak contributors
```

Sources/TokamakCore/Fiber/Layout/TextLayoutComputer.swift
```diff
@@ -0,0 +1,60 @@
+// Copyright 2021 Tokamak contributors
```
MaxDesiatov, line 1:
```suggestion
// Copyright 2022 Tokamak contributors
```

Sources/TokamakCore/Fiber/ViewGeometry.swift
```diff
@@ -0,0 +1,55 @@
+// Copyright 2021 Tokamak contributors
```
MaxDesiatov, line 1:
```suggestion
// Copyright 2022 Tokamak contributors
```

Sources/TokamakStaticHTML/Views/Text/Text.swift
```diff
@@ -158,6 +158,35 @@ extension Text: AnyHTML {
...
+    switch proxy.storage {
+    case let .verbatim(text):
+      innerHTML = proxy.environment.domTextSanitizer(text)
+    case let .segmentedText(segments):
+      innerHTML = segments
+        .map {
+          TextSpan(
+            content: proxy.environment.domTextSanitizer($0.0.rawText),
+            attributes: Self.attributes(
+              from: $0.1,
```
MaxDesiatov, line 174: nit: can we have labels added to this tuple type for clarity?
```suggestion
            content: proxy.environment.domTextSanitizer($0.storage.rawText),
            attributes: Self.attributes(
              from: $0.modifier,
```

Sources/TokamakStaticHTML/Views/Text/Text.swift
```diff
@@ -158,6 +158,35 @@ extension Text: AnyHTML {
...
+              environment: proxy.environment
+            )
+          )
+          .outerHTML(shouldSortAttributes: false, children: [])
+        }
+        .reduce("", +)
```
MaxDesiatov, line 180: I have a feeling that `reduce` with `+` may be slower than `reduce(into:)` with `Array.append`, but I haven't done any benchmarks. The thinking is that `+` causes heap allocations for newly created result strings, while `append` reuses the original string. Feel free to ignore this suggestion if you're not inclined to benchmark and micro-optimize this.

MaxDesiatov (commented):
As usual, you've done an amazing work here! 👏 I only have a couple of nits and a feeling that we need to document the layout algorithm more thoroughly. For example, a lot of articles on SwiftUI layout mention that parent proposes sizing for its children, which then can propose their own size in response, IIRC. Maybe we could have these details mentioned in doc comments of relevant types or inline comments where appropriate?

That is, it's hard for me to verify the correctness of the algorithm without having its description in plain words ready in the context of this code. But these documentation improvements may deserve their own separate PR.

Sources/TokamakCore/Environment/EnvironmentKey.swift
carson-katri: I added a doc comment for now. `EnvironmentalModifier` is a protocol available in SwiftUI that lets a user create a new `ViewModifier` using the environment. `_EnvironmentModifier` is a Tokamak protocol that lets a `View` or `ViewModifier` mutate the `EnvironmentValues`. The names are definitely confusing.

Sources/TokamakStaticHTML/Views/Text/Text.swift
carson-katri: I implemented this change to go from a `map` + `reduce("", +)` to a single `reduce(into:)`, but I haven't benchmarked it.

Sources/TokamakCore/Environment/EnvironmentKey.swift
MaxDesiatov: Should we keep the names as they are then?

carson-katri: I’d be fine renaming it, just don’t have an idea atm. `EnvironmentalModifier` needs to stay the same for compatibility though.

Sources/TokamakCore/Fiber/Layout/StackLayoutComputer.swift
carson-katri: Let me know if the new doc comment clears it up at all

carson-katri:
I would like to add some documentation in this PR, then its process can be more easily reviewed now rather than later

Sources/TokamakCore/Environment/EnvironmentKey.swift
MaxDesiatov: Yeah, let's keep it as is for now until we come up with a better name...

Sources/TokamakCore/Fiber/Layout/StackLayoutComputer.swift
```diff
@@ -0,0 +1,156 @@
...
+  let proposedSize: CGSize
+  let axis: Axis
+  let alignment: Alignment
+  let spacing: CGFloat
+  /// Will contain a value of `1` for the axis that takes the size of the largest child.
+  /// For instance, a `VStack` will provide `1` for the width because
+  /// it takes the width of its largest child.
+  ///
+  /// Typically flipped from the width/height of `fitAxis`
+  let maxAxis: CGSize
```
MaxDesiatov, line 30: It's a bit unclear to me though why this type is `CGSize` if it contains a value of `1` according to the comment. Is that `1` encoded in the size somehow? Why so?

MaxDesiatov: That is, if possible values are only `1` or `0`, shouldn't it be something like `(width: Bool, height: Bool)`?

Sources/TokamakCore/Fiber/FiberReconciler.swift
```diff
@@ -228,75 +186,219 @@ public final class FiberReconciler<Renderer: FiberRenderer> {
...
+                }
+              }
+            }
+            guard let parent = node.parent else { return }
+            // When we walk back to the root, exit
+            guard parent !== currentRoot.alternate else { return }
+            node = parent
+          }
+
+          // We also `size` and `position` when we reach the bottommost view that has a sibling.
```
MaxDesiatov, line 362:
```suggestion
          // We also request `size` and `position` when we reach the bottom-most view that has a sibling.
```

Sources/TokamakCore/Fiber/FiberReconciler.swift
```diff
@@ -228,75 +186,219 @@ public final class FiberReconciler<Renderer: FiberRenderer> {
...
+                // its parent.
+                mutations.insert(.remove(element: element, parent: parent), at: 0)
+              }
+              alternateSibling = alternateSibling?.sibling
+            }
+            // We `size` and `position` when we are walking back up the tree.
+            if reconciler.renderer.shouldLayout,
+               let fiber = node.fiber
+            {
+              // The elementParent proposed a size for this fiber on the way down.
```
MaxDesiatov, line 342:
```suggestion
              // The `elementParent` proposed a size for this fiber on the way down.
```

Sources/TokamakCore/Fiber/Layout/FrameLayoutComputer.swift
```diff
@@ -0,0 +1,74 @@
+// Copyright 2021 Tokamak contributors
```
MaxDesiatov, line 1: nit
```suggestion
// Copyright 2022 Tokamak contributors
```

Sources/TokamakCore/Fiber/Layout/RootLayoutComputer.swift
```diff
@@ -0,0 +1,44 @@
+// Copyright 2021 Tokamak contributors
```
MaxDesiatov, line 1:
```suggestion
// Copyright 2022 Tokamak contributors
```

Sources/TokamakCore/Fiber/Layout/ShrinkWrapLayoutComputer.swift
```diff
@@ -0,0 +1,46 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//  Created by Carson Katri on 5/24/22.
+//
+
+import Foundation
+
+/// A `LayoutComputer` that takes the size of its children.
```
MaxDesiatov, line 20: This comment is a bit unclear to me. Takes where? Maybe "shrinks to the size of its children" IIUC?

Sources/TokamakCore/Fiber/Layout/StackLayoutComputer.swift
```diff
@@ -0,0 +1,148 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//  Created by Carson Katri on 5/24/22.
+//
+
+import Foundation
+
+final class StackLayoutComputer: LayoutComputer {
```
MaxDesiatov, line 20: nit: missing doc comment

Sources/TokamakCore/Fiber/Layout/TextLayoutComputer.swift
```diff
@@ -0,0 +1,60 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//  Created by Carson Katri on 5/24/22.
+//
+
+import Foundation
+
+/// Measures the bounds of the `Text` with modifiers and wrapping inside the `proposedSize`.
```
MaxDesiatov, line 20: nit: if I understand the purpose of this class correctly. BTW: does `LayoutComputer` have to be a `class`? Could we use `structs` for these `LayoutComputer` types to avoid dynamic allocations?

```suggestion
/// Measures the bounds of the `Text` with modifiers and wraps it inside the `proposedSize`.
```

Sources/TokamakCore/Fiber/LayoutComputer.swift
```diff
@@ -17,9 +17,35 @@
...
+/// Sizes from `proposeSize` will be clamped, so it is safe to return negative numbers.
+protocol LayoutComputer: AnyObject {
   /// Will be called every time a child is evaluated.
   /// The calls will always be in order, and no more than one call will be made per child.
-  func proposeSize<V: View>(for child: V, at index: Int) -> CGSize
+  func proposeSize<V: View>(for child: V, at index: Int, in context: LayoutContext) -> CGSize
+  /// The child responds with their size and we place them relative to our origin.
+  func position(_ child: LayoutContext.Child, in context: LayoutContext) -> CGPoint
+  /// Request a size for ourself from our parent.
+  func requestSize(in context: LayoutContext) -> CGSize
```
MaxDesiatov, line 50: nit: I personally prefer a bit more spacing for easier reading
```suggestion

  /// The child responds with their size and we place them relative to our origin.
  func position(_ child: LayoutContext.Child, in context: LayoutContext) -> CGPoint

  /// Request a size for ourself from our parent.
  func requestSize(in context: LayoutContext) -> CGSize
```

Sources/TokamakCore/Fiber/ViewArguments.swift
```diff
@@ -46,18 +43,19 @@ final class EnvironmentBox {
...
+  init<V>(
     inputs: ViewInputs<V>,
     environment: EnvironmentValues? = nil,
     preferences: _PreferenceStore? = nil,
-    size: CGSize? = nil,
-    layoutComputer: LayoutComputer? = nil
+    layoutComputer: ((CGSize) -> LayoutComputer)? = nil
   ) {
-    // Only replace the EnvironmentBox when we change the environment. Otherwise the same box can be reused.
+    // Only replace the EnvironmentBox when we change the environment.
```
MaxDesiatov, line 52: nit
```suggestion
    // Only replace the `EnvironmentBox` when we change the environment.
```

Sources/TokamakCore/Fiber/ViewGeometry.swift
```diff
@@ -0,0 +1,55 @@
...
+// limitations under the License.
+//
+//  Created by Carson Katri on 2/17/22.
+//
+
+import Foundation
+
+public struct ViewGeometry: Equatable {
+  @_spi(TokamakCore) public let origin: ViewOrigin
+  @_spi(TokamakCore) public let dimensions: ViewDimensions
```
MaxDesiatov, line 22: nit: formatting tweaks for easier skimming through these declarations
```suggestion
  @_spi(TokamakCore) 
  public let origin: ViewOrigin

  @_spi(TokamakCore) 
  public let dimensions: ViewDimensions
```

Sources/TokamakCore/Fiber/ViewGeometry.swift
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/// The position of the `View` relative to its parent.
+public struct ViewOrigin: Equatable {
+  @_spi(TokamakCore) public let origin: CGPoint
+
+  @_spi(TokamakCore) public var x: CGFloat { origin.x }
+  @_spi(TokamakCore) public var y: CGFloat { origin.y }
```
MaxDesiatov, line 30: ditto

```suggestion
  @_spi(TokamakCore) 
  public let origin: CGPoint

  @_spi(TokamakCore) 
  public var x: CGFloat { origin.x }
  @_spi(TokamakCore) 
  public var y: CGFloat { origin.y }
```

Sources/TokamakDOM/DOMFiberRenderer.swift
```diff
@@ -15,6 +15,7 @@
 //  Created by Carson Katri on 2/7/22.
 //
 
+import Foundation
 import JavaScriptKit
 @_spi(TokamakCore) import TokamakCore
 @_spi(TokamakStaticHTML) import TokamakStaticHTML
```
MaxDesiatov, line 21: nit: making formatting consistent for `@_spi` attributes while we're at it 🙂
```suggestion
@_spi(TokamakCore) 
import TokamakCore
@_spi(TokamakStaticHTML) 
import TokamakStaticHTML
```

Sources/TokamakStaticHTML/Modifiers/ViewModifier.swift
```diff
@@ -44,3 +44,14 @@ extension _ZIndexModifier: DOMViewModifier {
     ["style": "z-index: \(index);"]
   }
 }
+
+@_spi(TokamakStaticHTML) extension ModifiedContent: HTMLConvertible where Content: View,
```
MaxDesiatov, line 48:
```suggestion
@_spi(TokamakStaticHTML)
extension ModifiedContent: HTMLConvertible where Content: View,
```

Sources/TokamakStaticHTML/StaticHTMLFiberRenderer.swift
```diff
@@ -15,7 +15,8 @@
 //  Created by Carson Katri on 2/6/22.
 //
 
-import TokamakCore
+import Foundation
+@_spi(TokamakCore) import TokamakCore
```
MaxDesiatov, line 19:
```suggestion
@_spi(TokamakCore) 
import TokamakCore
```

Sources/TokamakStaticHTML/Views/Text/Text.swift
MaxDesiatov: has that changed been pushed? I don't see it here in the diff

Sources/TokamakCore/Fiber/Fiber.swift
```diff
@@ -15,6 +15,8 @@
 //  Created by Carson Katri on 2/15/22.
 //
 
+import Foundation
+
 @_spi(TokamakCore) public extension FiberReconciler {
```
MaxDesiatov, line 20: nit
```suggestion
@_spi(TokamakCore) 
public extension FiberReconciler {
```

MaxDesiatov (commented):
A few more formatting and doc comments nits

Sources/TokamakDOM/DOMFiberRenderer.swift
carson-katri: Is it possible to configure this in swiftformat so it does it on commit?

Sources/TokamakStaticHTML/Views/Text/Text.swift
carson-katri: Ah, sorry. Only did it in one place.

Sources/TokamakDOM/DOMFiberRenderer.swift
MaxDesiatov: There's [`wrapAttributes` rule](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#wrapattributes), but I don't see options for `import` declarations there. Still would help with the rest of our code, we need `prev-line` value set for `--funcattributes`, `--typeattributes`, and `--varattributes`.

Sources/TokamakCore/Fiber/Layout/StackLayoutComputer.swift
MaxDesiatov: Great, thanks!

Sources/TokamakDOM/DOMFiberRenderer.swift
carson-katri: I could do this in a separate PR so we can reformat all the files without this PR getting too cluttered.

MaxDesiatov: No worries, I'll do that now.

Sources/TokamakCore/Fiber/AlignmentID.swift
```diff
@@ -17,7 +17,27 @@
...
+///           return 10.0
+///         }
+///       }
+///       public static let myAlignmentGuide = Self(MyAlignmentGuide.self)
+///     }
+///
+/// Which you can then use with the `alignmentGuide` modifier:
+///
+///     Text("Hello, world!")
+///       .alignmentGuide(.myAlignmentGuide) { $0[.trailing] - 3 }
```
MaxDesiatov, line 37: Hm, to be honest I've never used `AlignmentID` before. Just tried this code in real Swift UI and it doesn't have any impact on `Text` positioning. Is there some context in which `.alignmentGuide` works? I'm just having this plain `Text` with `.alignmentGuide` modifier with a custom AG. Tried both in a stack and as a root view, same thing, no effect.

carson-katri: I have used them before, so this example is probably just bad. I didn't test it out, I'll come up with a working sample.

Sources/TokamakCore/Fiber/Fiber.swift
```diff
@@ -134,7 +144,14 @@
       if let element = element {
         self.element = element
       } else if Renderer.isPrimitive(view) {
-        self.element = .init(from: .init(from: view))
+        self
+          .element =
+          .init(from: .init(from: view, shouldLayout: reconciler?.renderer.shouldLayout ?? false))
```
MaxDesiatov, line 149: nit: weird formatting?
```suggestion
        self.element =
          .init(from: .init(from: view, shouldLayout: reconciler?.renderer.shouldLayout ?? false))
```

Sources/TokamakCore/Fiber/AlignmentID.swift
carson-katri: I updated the sample to show one scenario they could be used. You can use the custom guide as the alignment for a `VStack`. Then the `alignmentGuide` modifier will have an effect if you specify the same `AlignmentID` as the one to modify, because that's the one the VStack layout is reading.

We don't have the `alignmentGuide` modifier yet, but it should work the same once we implement it, and `AlignmentID` is used internally for all the builtin alignments.

Sources/TokamakCore/Fiber/Fiber.swift
```diff
@@ -134,7 +144,14 @@
...
+      }
+
+      // Only specify an elementIndex if we have an element.
```
MaxDesiatov, line 152:
```suggestion
      // Only specify an `elementIndex` if we have an element.
```

PR #475, opened 2022-04-05
Build and test with SwiftWasm 5.6 on CI

MaxDesiatov:
Had to drop support for Swift 5.4/5.5 and macOS 5.6 jobs, see https://github.com/TokamakUI/Tokamak/pull/475#issuecomment-1092662828 for more details.

Also applied a few formatting changes with the latest SwiftFormat.


MaxDesiatov:
I'm seeing a validation issue in `TokamakDemo.wasm` binaries built with SwiftWasm 5.6:

```
% ../wasminspect/target/debug/wasminspect ./.build/debug/TokamakDemo.wasm
[2022-04-05T10:19:11Z WARN  wasminspect_debugger::debugger] type mismatch: expected i32 but nothing on stack (at offset 22347435)
type mismatch: expected i32 but nothing on stack (at offset 22347435)
```

Broken binary produced by SwiftWasm 5.6 attached. It's also reproducible with `carton` 0.13.0 on this branch with `carton dev --product TokamakDemo `.

[TokamakDemo.wasm.zip](https://github.com/TokamakUI/Tokamak/files/8417178/TokamakDemo.wasm.zip)

cc @kateinoigakukun

kateinoigakukun:
I guess our WasmTransformer does something wrong...

Package.swift
```diff
@@ -1,32 +1,7 @@
-// swift-tools-version:5.4
+// swift-tools-version:5.6
```
kateinoigakukun, line 1: Why need to update tools version? Can we keep `Package@5.4.swift` for older toolchains?

MaxDesiatov:
I had to drop SwiftWasm 5.4/5.5, as those are incompatible with new package manifest format parsed by latest `carton`. Also makes sense to raise the requirement in our `Package.swift` to 5.6 to prevent users from stumbling upon that issue.

As a consequence, I had to disable `macos-11` jobs, as `macos-12` is not generally available yet on GHA, but they're required for Xcode 13.3 and Swift 5.6. I've submitted a request for `macos-12` for TokamakUI org, but it hasn't been granted yet.

In a seemingly unrelated way GTK Swift 5.6 jobs are crashing on Linux with this error:

```
1.	Swift version 5.6 (swift-5.6-RELEASE)
2.	Compiling with the current language version
3.	While evaluating request TypeCheckSourceFileRequest(source_file "/home/runner/work/Tokamak/Tokamak/Sources/TokamakCore/ViewTraits/Transition/Transition.swift")
4.	While type-checking '_AnyTransitionModifier' (at /home/runner/work/Tokamak/Tokamak/Sources/TokamakCore/ViewTraits/Transition/Transition.swift:61:8)
5.	While type-checking 'isActive' (at /home/runner/work/Tokamak/Tokamak/Sources/TokamakCore/ViewTraits/Transition/Transition.swift:64:7)
6.	While evaluating request CheckRedeclarationRequest(TokamakCore.(file)._AnyTransitionModifier.isActive@/home/runner/work/Tokamak/Tokamak/Sources/TokamakCore/ViewTraits/Transition/Transition.swift:64:7)
7.	While building rewrite system for generic signature <τ_0_0 where τ_0_0 : _AnyTransitionModifier>
```

I'm disabling those jobs as well until those compiler crashes are fixed.

Package.swift
MaxDesiatov: Let me double-check this once again... I've tried this once and 5.4/5.5 didn't really work with latest `carton`

MaxDesiatov:
I managed to re-enable GTK Linux builds, turns out the compiler crash is fixed in nightly builds.

carson-katri:
Awesome, that’s the same error I was seeing on the fiber branch with the Linux builds, and was stumped.

ezraberch:
Dropping support for Swift <5.6 should mean we can get rid of all of the references to earlier versions throughout the codebase.

i.e.

`#if compiler(>=5.5) && (canImport(Concurrency) || canImport(_Concurrency))`

MaxDesiatov:
Thanks, that's a great point. I've addressed it in the latest commit.

.github/workflows/codecov.yml
```diff
@@ -1,4 +1,4 @@
-
+# Disable until crashes with Swift 5.6 are fixed.
```
j-f1, line 1: This comment can be removed, right?

MaxDesiatov, line 1:
```suggestion
```

j-f1 (approved):

Issue #269, opened 2020-08-27
Allow modifying the `<head>` DOM element


MaxDesiatov:
The main use case would be in modifying the `<title>` value and adding styles/scripts declaratively with support for the `TokamakStaticHTML` renderer. I'm more inclined towards adding the `HTMLHead` type conforming to `Scene`, to which one could add `HTML` views with whatever content they want.

carson-katri:
So it'd be like:
```swift
struct MyApp : App {
}
```

carson-katri:
So it'd be like this?:
```swift
struct MyApp : App {
  var body: some Scene {
    HTMLHead {
      HTML("meta", [...])
    }
    WindowGroup("MyApp") {
      ContentView()
    }
  }
}
```
If so, I'm all for it being a `Scene`.

MaxDesiatov:
Yes, exactly. If there's a need for a given view (or `Router`) to adjust anything, a given `App` could expose it as an environment object for example.

j-f1:
> most of the time it would be the `<title>` content

Since macOS already has the concept of a title for a window, shouldn’t we implement that API to enable setting the title?

MaxDesiatov:
That's a great point, but I previously haven't written any SwiftUI apps for macOS, so I don't even know how that's supposed to work to propose anything concrete on that front. I obviously would prefer using that if it fits well for our purposes.

carson-katri:
We have it, it's the String/Text argument passed to the WindowGroup initializer:
`WindowGroup("My Site Title") { ... }`

MaxDesiatov:
Right, I started thinking of `<head>` because I tried to imagine how `<title>` changes could be integrated with the new router, I'm not sure it would be that easy if the only way to set a title is via the `WindowGroup` argument. Unless we add an explicit title binding initializer argument to the `Router` type?

carson-katri:
Ah, ok. So in SwiftUI I think the only way to do it before macOS 11 was via the AppDelegate. So using the `HTMLHead` would probably be best in that case.

MaxDesiatov:
IDK, with the environment object I mentioned above the title can be dynamic already if only the router supported title bindings:

```swift
struct MyApp : App {
  @State var title: String

  var body: some Scene {
    WindowGroup(title) {
      Router(title: $title) {
        // more routes here, but unclear yet how they adjust the title
      }
    }
  }
}
```

[...]

PR #483, opened 2022-05-19
Support meta tags in StaticHTMLRenderer

AndrewBarba:
This PR adds the ability to control `<head>` tags using a new `HTMLTitle` view and `HTMLMeta` view. Taking inspiration from Next.js `<NextHead>`, you can use these views anywhere in your view hierarchy and they will be hoisted to the top `<head>` section of the html.

Use as a view:

```swift
var body: some View {
  VStack {
    ...
    HTMLTitle("Hello, Tokamak")
    HTMLMeta(charset: "utf-8")
    ...
  }
}
```

Use as a view modifier:

```swift
var body: some View {
  VStack {
    ...
  }
  .htmlTitle("Hello, Tokamak")
  .htmlMeta(charset: "utf-8")
}
```

And the resulting html (no matter where these are used in your view hierarchy):

```html
<html>
  <head>
    <title>Hello, Tokamak</title>
    <meta charset="utf-8">
  </head>
  <body>
    ...
  </body>
</html>
```

### DIFF
--- a/Sources/TokamakCore/StackReconciler.swift
+++ b/Sources/TokamakCore/StackReconciler.swift
@@ -58,6 +58,12 @@ public final class StackReconciler<R: Renderer> {
    */
   public let rootTarget: R.TargetType
 
+  /** A root renderer's main preference store
+   */
+  public var preferenceStore: _PreferenceStore {
+    rootElement.preferenceStore
+  }
+
   /** A root of the mounted elements tree to which all other mounted elements are attached to.
    */
   private let rootElement: MountedElement<R>
--- a/Sources/TokamakCore/StackReconciler.swift
+++ b/Sources/TokamakCore/StackReconciler.swift
@@ -58,7 +58,7 @@ public final class StackReconciler<R: Renderer> {
    */
   public let rootTarget: R.TargetType
 
-  /** A root renderer's main preference store
+  /** A root renderer's main preference store.
    */
   public var preferenceStore: _PreferenceStore {
     rootElement.preferenceStore
--- a/Sources/TokamakStaticHTML/Preferences/Preferences.swift
+++ b/Sources/TokamakStaticHTML/Preferences/Preferences.swift
@@ -29,6 +29,6 @@ struct HTMLMetaPreferenceKey: PreferenceKey {
     static var defaultValue: [_HTMLMeta] = []
 
     static func reduce(value: inout [_HTMLMeta], nextValue: () -> [_HTMLMeta]) {
-        value = value + nextValue()
+        value += nextValue()
     }
 }
--- a/Sources/TokamakStaticHTML/Preferences/Preferences.swift
+++ b/Sources/TokamakStaticHTML/Preferences/Preferences.swift
@@ -26,16 +26,9 @@ struct HTMLTitlePreferenceKey: PreferenceKey {
 }
 
 struct HTMLMetaPreferenceKey: PreferenceKey {
-    internal enum HTMLMeta: Equatable, Hashable {
-        case charset(_ charset: String)
-        case name(_ name: String, content: String)
-        case property(_ property: String, content: String)
-        case httpEquiv(_ httpEquiv: String, content: String)
-    }
+    static var defaultValue: [_HTMLMeta] = []
 
-    static var defaultValue: [HTMLMeta] = []
-
-    static func reduce(value: inout [HTMLMeta], nextValue: () -> [HTMLMeta]) {
+    static func reduce(value: inout [_HTMLMeta], nextValue: () -> [_HTMLMeta]) {
         value = value + nextValue()
     }
 }
--- a/Sources/TokamakStaticHTML/Preferences/Preferences.swift
+++ b/Sources/TokamakStaticHTML/Preferences/Preferences.swift
@@ -17,18 +17,21 @@
 
 import TokamakCore
 
-struct HTMLTitlePreferenceKey: PreferenceKey {
-    static var defaultValue: String = ""
+public struct HTMLTitlePreferenceKey: PreferenceKey {
+    public static var defaultValue: String = ""
 
-    static func reduce(value: inout String, nextValue: () -> String) {
+    public static func reduce(value: inout String, nextValue: () -> String) {
         value = nextValue()
     }
 }
 
-struct HTMLMetaPreferenceKey: PreferenceKey {
-    static var defaultValue: [_HTMLMeta] = []
+public struct HTMLMetaPreferenceKey: PreferenceKey {
+    public static var defaultValue: [HTMLMeta.MetaTag] = []
 
-    static func reduce(value: inout [_HTMLMeta], nextValue: () -> [_HTMLMeta]) {
+    public static func reduce(
+        value: inout [HTMLMeta.MetaTag],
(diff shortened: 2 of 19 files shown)


carson-katri:
There was previously some discussion in #269  around designing a way to modify the head through a Scene similar to the `Settings` scene in SwiftUI:

```swift
struct MyApp : App {
  var body: some Scene {
    Head {
      Meta(…)
    }
    WindowGroup(…) {
      ContentView()
    }
  }
}
```

MaxDesiatov:
@AndrewBarba would do you think about the suggested `Head`/`Meta` API that uses types conforming to `Scene`? I also think that we should support the `property` attribute for OpenGraph meta tags.

MaxDesiatov:
Yeah, that would require bigger changes to the renderer though. Let me know if you'd like to implement them yourself or have any questions. I can pick this up next week otherwise.

AndrewBarba:
This kind of makes sense. I would expect a new initializer that takes in a Scene which defines the entire HTML structure - including Meta tags and titles? And then the existing init that just takes in a View would wrap that View with this new base HTML structure. I have so confusion around `HTMLTarget`, if I make a full htmlTree as the target then all the children seem to be overwritten. Im not sure what the main use of the `HTMLTarget` struct actually is to be honest.

MaxDesiatov:
> I would expect a new initializer that takes in a Scene which defines the entire HTML structure - including Meta tags and titles?

No need for a new initializer, I don't think one would pass `Scene` values instead of `App` values, and already have an initializer for the latter.

> I have some confusion around `HTMLTarget`, if I make a full htmlTree as the target then all the children seem to be overwritten. Im not sure what the main use of the `HTMLTarget` struct actually is to be honest.

Sorry about the confusion. `HTMLTarget` is a class that represents tree state the reconciler updates when needed. I misremembered the logic here, you're right, initial `HTMLTarget` value will always be overwritten. You'll have to modify `extension WindowGroup: SceneDeferredToRenderer` to wrap `view` in `<head>` and `<body>` tags instead.

AndrewBarba:
@MaxDesiatov @carson-katri Check it out, I refactored PR to use preference keys so now you can drop `Title` and `Meta` anywhere in your view hierarchy just like next.js and its hoisted to the top. I'm pretty happy with end result but let me know if you want me to add scripts, styles or anything else. Also very open to feedback on how I organized the files.

```swift
var body: some View {
  VStack {
    ...
    Title("Hello, Tokamak")
    Meta(charset: "utf-8")
    ...
  }
}
```

You can see the result working here: https://tokamak.app.swift.cloud

AndrewBarba:
I agree, I think prefixing makes more sense. I made this change and updated the PR description to show the correct usage. I added usage for both views and view modifiers. SwiftUI on iOS uses a lot of modifiers, ex. `navigationBarTitle`, so I think it's important to support both.

Sources/TokamakCore/StackReconciler.swift
```diff
@@ -58,6 +58,12 @@ public final class StackReconciler<R: Renderer> {
    */
   public let rootTarget: R.TargetType
 
+  /** A root renderer's main preference store
```
MaxDesiatov, line 61: nit: comment formatting 
```suggestion
  /** A root renderer's main preference store.
```

Sources/TokamakStaticHTML/Preferences/Preferences.swift
```diff
@@ -0,0 +1,41 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//  Created by Andrew Barba on 5/20/22.
+//
+
+import TokamakCore
+
+internal struct HTMLTitlePreferenceKey: PreferenceKey {
```
MaxDesiatov, line 20: We don't use `internal` in our codebase as it's the default visibility and is redundant.

```suggestion
struct HTMLTitlePreferenceKey: PreferenceKey {
```

MaxDesiatov (commented):
There are a few nits to pick, but overall this makes sense. One more question: how does this behave if there are multiple `Title` views in the same tree? Whichever comes last is applied I guess? Would you be able to add a snapshot test that confirms this? Thanks!

carson-katri (changes requested):
Overall it looks good to me, just needs a couple tests. I think they'd go in `TokamakStaticHTMLTests/HTMLTests.swift`

carson-katri:
Could you add a test for `HTMLMeta` too?

AndrewBarba:
Definitely - working on them now

AndrewBarba:
I pushed a test that uses 4 meta tags but only the last one ends up in the resulting html. Currently trying to debug this, I'm pretty sure my implementation is correct so I wonder if there's an issue with the preference store.

AndrewBarba:
Okay the main issue here is that every mount target gets its own preference store and then we are trying to merge upwards anytime a child store changes. This is problematic because the merge only saves the value of the child and doesn't maintain any values in the parent for the same key.

I was able to fix this by:

1. Change `_PreferenceStore` to a class
2. Initialize `MountedElement.preferenceStore` to the parents store or create a new one

I can push this up if you think it makes sense, otherwise I'm open to other approaches.

MaxDesiatov:
IIRC @carson-katri is the author of preference store code, so I'll defer to him on this matter if he has a moment to review this.

carson-katri:
So does this make the child take all the values of the parent's preference store? If so, I don't think this behavior is correct as preference values flow up the tree not down. Similar to how an environment value from a child is not set on the parent, a preference value from a parent is not set on the child.

AndrewBarba:
Ah yes you are right. Don't think we have tests for this so I will add them. 

So I think the approach to solve that is add a weak parent store to the _PreferenceStore. Each mount target will go back to getting its own store but the pref store will be updated to modify its parent store anytime one of its keys changes. This is still a better solution to merge, because the issue with merge is it doesn't know which keys changed, and so even with some fancy generic work to deep merge keys upwards, it was over adding values to the parent store because it was doing a full merge on ever didSet. If we control the upstream changes to only happen on inserting a key this should be more reliable.

AndrewBarba:
Okay just pushed a test that I think correctly covers what you described.

carson-katri (approved):
That looks good to me, thanks!

MaxDesiatov (changes requested):
Overall this is great, one small change request: could you run SwiftFormat, and potentially have a pre-commit hook installed as described in [the "Coding Style" section of `CONTRIBUTING.md`](https://github.com/TokamakUI/Tokamak/blob/main/CONTRIBUTING.md#coding-style)? Some of the newly added files use 4 spaces for indentation, while we use 2 spaces everywhere in the project.

Thanks!

AndrewBarba:
Yes definitely. Xcode is so annoying with the indents, I thought I caught them all. Will get this fixed

MaxDesiatov:
Don't worry about catching them in Xcode manually. When I edit in Xcode I just write some messy inconsistent indentation, and it's all reformatted with pre-commit on every commit automatically 🙂

AndrewBarba:
Yeah wow - this is so much better. My bad for not reading that code style doc earlier

MaxDesiatov (approved):
Lovely, thanks!

PR #485, opened 2022-05-23
Add `_PaddingLayout` support to layout engine

carson-katri:
This adds support for the `padding` modifier to the custom layout engine.


Sources/TokamakStaticHTML/Modifiers/LayoutModifiers.swift
```diff
@@ -144,6 +144,14 @@ extension _PaddingLayout: DOMViewModifier {
   }
 }
 
+@_spi(TokamakStaticHTML) extension _PaddingLayout: HTMLConvertible {
```
MaxDesiatov, line 147:
```suggestion
@_spi(TokamakStaticHTML) 
extension _PaddingLayout: HTMLConvertible {
```

Sources/TokamakCore/Fiber/Layout/PaddingLayoutComputer.swift
```diff
@@ -0,0 +1,79 @@
...
+      top: edges.contains(.top) ? insets.top : 0,
+      leading: edges.contains(.leading) ? insets.leading : 0,
+      bottom: edges.contains(.bottom) ? insets.bottom : 0,
+      trailing: edges.contains(.trailing) ? insets.trailing : 0
+    )
+  }
+}
+
+/// A `LayoutComputer` that fits to its children then adds padding.
+final class PaddingLayoutComputer: LayoutComputer {
```
MaxDesiatov, line 32: I think I had a similar comment before, but got no clarity on this yet: are types conforming to `LayoutComputer` supposed to be classes or structs? If structs are fine, let's avoid classes without an explicit need for them, which helps us avoid slower allocations on the heap.

Sources/TokamakCore/Fiber/Layout/PaddingLayoutComputer.swift
```diff
@@ -0,0 +1,79 @@
...
+  }
+}
+
+public extension _PaddingLayout {
+  static func _makeView(_ inputs: ViewInputs<Self>) -> ViewOutputs {
+    .init(inputs: inputs, layoutComputer: { PaddingLayoutComputer(
+      proposedSize: $0,
+      edges: inputs.content.edges,
+      insets: inputs.content.insets
+    ) })
```
MaxDesiatov, line 77: nit: is this formatting applied by SwiftFormat?

Sources/TokamakCore/Fiber/Layout/PaddingLayoutComputer.swift
j-f1: …and if they should be classes, then the `LayoutComputer` protocol should inherit from `AnyObject` to ensure that.

Sources/TokamakCore/Fiber/Layout/PaddingLayoutComputer.swift
carson-katri: Must be, I don't remember doing this.

Sources/TokamakCore/Fiber/Layout/PaddingLayoutComputer.swift
carson-katri: They should be structs now, I has changed that before merging the layout branch, but had made this PR before that change was made.

MaxDesiatov (approved):
Thanks 🙏

PR #486, opened 2022-05-24
Add html5 doctype to static renderer

AndrewBarba:
- Add `<!DOCTYPE html>` to the top of static rendered pages

PR #484, opened 2022-05-23
Reenable macOS CI builds, using macOS 12 and Xcode 13.4

ezraberch:
Now that the macOS 12 image is available, the CI builds can be reenabled.

PR #487, opened 2022-05-24
Revert "Add html5 doctype to static renderer (#486)"

MaxDesiatov:
This reverts commit 3081f5521a7bfc10e4b3087f5744887a1c7f4947 which causes failing layout snapshots we've recently turned back on in https://github.com/TokamakUI/Tokamak/pull/484.


ezraberch:
The rest of the failing tests are from #483. Those are caused by an extra blank line after the title tag.

MaxDesiatov:
CC @AndrewBarba

ezraberch (approved):

Issue #490, opened 2022-05-29
Binary code size benchmarks


MaxDesiatov:
It would be interesting to see how size of simple `Text("Hello, World")` example binary changes with changes made to Tokamak. We should consider gathering these metrics on CI and introducing a GitHub Actions job that can measure this.

PR #496, opened 2022-05-31
Add file size to benchmark script

ezraberch:
As suggested in #490, this PR adds a file size display to CI. The benchmark script will now display the file size of both benchmarks.

Additionally, the TokamakDemo build in CI was still disabled. This has been fixed.

### DIFF
--- a/benchmark.sh
+++ b/benchmark.sh
@@ -3,6 +3,8 @@
 set -eux
 
 swift build -c release --product TokamakCoreBenchmark
+stat -f "TokamakCoreBenchmark: %z bytes" ./.build/release/TokamakCoreBenchmark
 ./.build/release/TokamakCoreBenchmark
 swift build -c release --product TokamakStaticHTMLBenchmark
+stat -f "TokamakStaticHTMLBenchmark: %z bytes" ./.build/release/TokamakStaticHTMLBenchmark
 ./.build/release/TokamakStaticHTMLBenchmark
--- a/benchmark.sh
+++ b/benchmark.sh
@@ -3,8 +3,8 @@
 set -eux
 
 swift build -c release --product TokamakCoreBenchmark
-stat -f "TokamakCoreBenchmark: %z bytes" ./.build/release/TokamakCoreBenchmark
+stat -f "::warning file=Sources/TokamakCoreBenchmark/main.swift,line=1,col=1::TokamakCoreBenchmark is %z bytes." ./.build/release/TokamakCoreBenchmark
 ./.build/release/TokamakCoreBenchmark
 swift build -c release --product TokamakStaticHTMLBenchmark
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
How hard would it be instead of logging this from `benchmark.sh` to post it as a comment on PR being tested? Looks like there's a seemingly simple GHA script solution for posting the actual comment, the question is only about passing this data to the comment  https://stackoverflow.com/a/64126737/442427

To avoid spamming every PR we could only post a comment when there's a significant enough change in size, say 5%+. WDYT?

carson-katri:
Also, should/can we track the size of the SwiftWasm build as well?

MaxDesiatov:
Yes, size of `.wasm` binaries produced by `carton bundle` is the main thing I'm personally interested in.

ezraberch:
> How hard would it be instead of logging this from `benchmark.sh` to post it as a comment on PR being tested?

That's doable, but wouldn't an annotation (like what codecov does) be more appropriate?

> Also, should/can we track the size of the SwiftWasm build as well?

Are you referring to this from the bundle check?

```
Right after building the main binary size is 31.67 MB

After stripping debug info the main binary size is 20.61 MB
```

MaxDesiatov:
> That's doable, but wouldn't an annotation (like what codecov does) be more appropriate?
I've been thinking about that, but the problem with annotations is that they're easy to miss. 

Maybe we could start with annotation for now if that's easier to implement.

> Are you referring to this from the bundle check?
> 
> ```
> Right after building the main binary size is 31.67 MB
> 
> After stripping debug info the main binary size is 20.61 MB
> ```

Yes, but maybe it's easier to avoid parsing the output from `carton bundle` and to just run `du -h Bundle/*.wasm` as a step in `swiftwasm_bundle_5_6` job.

benchmark.sh
```diff
@@ -3,6 +3,8 @@
 set -eux
 
 swift build -c release --product TokamakCoreBenchmark
+stat -f "::warning file=Sources/TokamakCoreBenchmark/main.swift,line=1,col=1::TokamakCoreBenchmark is %z bytes." ./.build/release/TokamakCoreBenchmark
 ./.build/release/TokamakCoreBenchmark
 swift build -c release --product TokamakStaticHTMLBenchmark
+stat -f "::warning file=Sources/TokamakStaticHTMLBenchmark/main.swift,line=1,col=1::TokamakStaticHTMLBenchmark is %z bytes." ./.build/release/TokamakStaticHTMLBenchmark
```
MaxDesiatov, line 9: Can we have this in a more readable format, similar to what `du -h` prints. Or maybe we can use `du -h` directly here?

```suggestion
stat -f "::warning file=Sources/TokamakStaticHTMLBenchmark/main.swift,line=1,col=1::TokamakStaticHTMLBenchmark is %z bytes." ./.build/release/TokamakStaticHTMLBenchmark
```

PR #497, opened 2022-06-07
Implement `AnyShape`

filip-sakel:
This PR is a starting point for catching up with SwiftUI 4. There are still some unknowns, like whether the behavior of adding two _AnyAnimatableData instances that erase different types should crash (which is marked with a TODO), but the API likely won't change that much.

### DIFF
--- a/Sources/TokamakCore/Animation/_AnyAnimatableData.swift
+++ b/Sources/TokamakCore/Animation/_AnyAnimatableData.swift
@@ -1,0 +1,165 @@
+// Copyright 2022 Tokamak contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


Sources/TokamakCore/Animation/_AnyAnimatableData.swift
```diff
@@ -0,0 +1,165 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// A type-eraser for `VectorArithmetic`.
+public struct _AnyAnimatableData: VectorArithmetic {
+  private var box: _AnyAnimatableDataBox?
```
filip-sakel, line 17: SwiftUI implements this using an `_AnyAnimatableDataVTable` class, but I didn't find a reason for having this implementation complexity. From some quick testing, I found that `AnyShape(Circle()).animatableData` just have a `VTable<Circle>` implementation, while `_AnyAnimatableDataVTable.zero` uses `ZeroVTable`. This implementation, however, should be the same as it is with SwiftUI.

carson-katri (commented):
LGTM overall! Just had one question

carson-katri (approved):

PR #498, opened 2022-06-16
Add `Layout` protocol for FiberReconciler

carson-katri:
This adds support for the new `Layout` protocol introduced in iOS 16 and aligned releases.

The `FiberReconciler`'s layout code has been refactored to use this protocol in place of the previous `LayoutComputer` implementation. In order to match the behavior of native SwiftUI, I split the reconciling and layout passes into two separate loops. The reason for this change is that all of the `LayoutSubview`s must be collected prior to performing any layout computations. However, the previous `LayoutComputer` approach built the size incrementally as the `View`s appeared.

`VStack`/`HStack`, and other `LayoutComputer`s have been rewritten to use the `Layout` protocol. I specifically worked on getting the stacks to match SwiftUI as closely as possible, other layouts may need tweaking.

### TODOs:
- [x] Documentation on layout pass
- [x] Compare benchmarks with previous layout code
- [x] Maybe: Refactor FiberReconciler.swift to split the reconciliation and layout loops into separate files
- [x] Cleanup filenames referencing`LayoutComputer`
- [x] Add tests for `Layout` protocol (possibly test against the computed frames)
- [x] Allow renderers and primitive views to customize `ViewSpacing` (for instance, `Text` seems to prefer no spacing on the top/bottom, but default spacing on the leading/trailing edges)
- [x] Background layout does not respect alignment
- [x] `Layout.Cache` does not actually cache between runs
- [x] Support `LayoutValueKey`


carson-katri:
Here are the results of the TokamakCoreBenchmark with layout enabled:

### `Layout` protocol
```
name                             time      std        iterations
----------------------------------------------------------------
update wide (StackReconciler)    46.404 ms ±   1.92 %         30
update wide (FiberReconciler)    42.673 ms ±   3.23 %         33
update narrow (StackReconciler)  45.874 ms ±   2.04 %         31
update narrow (FiberReconciler)  42.407 ms ±   1.84 %         33
update deep (StackReconciler)    17.447 ms ±   2.48 %         80
update deep (FiberReconciler)     7.656 ms ±   8.38 %        177
update shallow (StackReconciler)  8.899 ms ±   1.93 %        157
update shallow (FiberReconciler)  4.995 ms ±   4.32 %        277
```

### `LayoutComputer` implementation
```
name                             time      std        iterations
----------------------------------------------------------------
update wide (StackReconciler)    46.891 ms ±   1.67 %         30
update wide (FiberReconciler)    33.054 ms ±   1.82 %         42
update narrow (StackReconciler)  46.437 ms ±   2.98 %         30
update narrow (FiberReconciler)  33.150 ms ±   3.61 %         42
update deep (StackReconciler)    17.587 ms ±   1.54 %         80
update deep (FiberReconciler)     6.135 ms ±   3.47 %        227
update shallow (StackReconciler)  8.952 ms ±   1.56 %        156
update shallow (FiberReconciler)  3.854 ms ±   3.17 %        363
```

carson-katri:
Graphs! I wanted to visualize the performance of these different methods to see how they compare more visually (easier on the eyes than a bunch of nanosecond numbers 😅).

> x-axis is # of views, y-axis is time in ms to update (not the first render).

<details>
<summary>`ForEach` Test</summary>

Rendering from 1-5000 views in a `VStack` with `ForEach`. This tests an update, not the first render.
Here is the View used to test:
```swift
struct TestView: View {
  let items: Int
  @State
  var update = -1

  var body: some View {
    VStack {
      ForEach(0..<items) {
        if update == $0 {
          Text("Updated")
        } else {
          Text("\($0)")
        }
      }
      Button("Update") {
        update = items - 1
      }
    }
  }
}
```

Screen Shot 2022-06-17 at 6 22 17 PM
</details>

<details>
<summary>`ForEach` Test (first 30 values)</summary>

Screen Shot 2022-06-17 at 6 23 00 PM

</details>

<details>
<summary>`ForEach` Test (first render)</summary>
This is the same as the `ForEach` test, but it only tests the first render, not an update.

Screen Shot 2022-06-17 at 6 47 55 PM

</details>

<details>
<summary>`RecursiveView` Test</summary>
This test doesn't stress the layout engine because the number of visible Views does not change, it only renders a `Text` at the end of the recursive chain. This also tests an update, not the first render.

Screen Shot 2022-06-17 at 6 39 01 PM

</details>

carson-katri:
I added a few things that have optimized the Layout protocol implementation for the "`ForEach` Test" from above. Specifically, I attempted to optimize array capacity by having `View`s that know their child counts implement a `_viewChildrenCount` method, which gives us the information needed to `reserveCapacity` on the `LayoutSubviews` array.

The cache is now persisted as well, although currently I don't think any built-in `Layout` implementations would benefit much from this.

Here is the new graph:

Screen Shot 2022-06-17 at 10 59 49 PM

carson-katri:
I've added a new `TokamakLayoutTests` target, which compares native SwiftUI to custom layout implementations via image.

I only use grayscale colors in those tests because the RGB colors don't quite match between the images, and we really just want to test the actual view geometry.

carson-katri:
Not sure how accurate this is, but I ran the same `ForEach` benchmark on native SwiftUI in the iOS 16 simulator, and it was actually slower than this implementation and crashed before it got to 10,000 views.

Screen Shot 2022-06-23 at 3 46 01 PM

This tests the first render for a SwiftUI view though, not an update like the other tests. I think an update will be significantly faster than this, but I'm not sure how to benchmark that. I used the below snippet to run the test in SwiftUI:

```swift
for i in stride(from: 1, to: 10000, by: 100) {
  let start = DispatchTime.now()
  let controller = UIHostingController(rootView: TestView(items: i))
  _ = controller.view.intrinsicContentSize // Access the size to trigger SwiftUI layout.
  let end = DispatchTime.now()
  graph.append(end.uptimeNanoseconds - start.uptimeNanoseconds)
}
```

MaxDesiatov:
Sorry for the delay with the review. I have it on my list and will get to it during this weekend.

carson-katri:
Here's my attempt at measuring an update (not a first render) in SwiftUI. This is a more accurate comparison, but still not perfect:

Screen Shot 2022-06-24 at 10 58 45 AM

Sources/TokamakCore/Fiber/Layout/Layout.swift
```diff
@@ -0,0 +1,464 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//  Created by Carson Katri on 2/16/22.
+//
+
+import Foundation
+
+public protocol _AnyLayout {
```
MaxDesiatov, line 20: This needs a doc comment

Sources/TokamakCore/Fiber/Layout/Layout.swift
```diff
@@ -0,0 +1,464 @@
...
+  func _erased() -> AnyLayout
+}
+
+/// A type that participates in the layout pass.
+///
+/// Any `View` or `Scene` that implements this protocol will be used to computed layout in
+/// a `FiberRenderer` with `useDynamicLayout` enabled.
```
MaxDesiatov, line 27: nit

```suggestion
/// Any `View` or `Scene` that implements this protocol will be used to compute layout in
/// a `FiberRenderer` with `useDynamicLayout` set to `true`.
```

Sources/TokamakCore/Fiber/Layout/Layout.swift
```diff
@@ -0,0 +1,464 @@
...
+    cache: inout ()
+  ) {
+    for subview in subviews {
+      subview.place(at: bounds.origin, proposal: proposal)
+    }
+  }
+}
+
+@usableFromInline
+protocol AnyLayoutBox {
```
MaxDesiatov, line 236: this needs a doc comment

Package.swift
```diff
@@ -195,6 +195,18 @@ let package = Package(
       name: "TokamakTestRenderer",
       dependencies: ["TokamakCore"]
     ),
+    .testTarget(
+      name: "TokamakLayoutTests",
```
MaxDesiatov, line 199: Are these new tests running on CI? I don't see any changes to GHA workflows in this diff.

Sources/TokamakCore/Fiber/Fiber.swift
```diff
@@ -54,11 +54,15 @@ public extension FiberReconciler {
     /// all stored properties be set before using.
     /// `outputs` is guaranteed to be set in the initializer.
     var outputs: ViewOutputs!
+    /// The erased `Layout` to use for this content.
```
MaxDesiatov, line 57: nit: while this file is edited, could we add a newline between a previous property and a doc comment of a following property for readability? There are obviously more places in this file where this change could be applied.
```suggestion

    /// The erased `Layout` to use for this content.
```

Sources/TokamakCore/Fiber/FiberReconciler.swift
```diff
@@ -31,6 +31,10 @@ public final class FiberReconciler<Renderer: FiberRenderer> {
   /// The `FiberRenderer` used to create and update the `Element`s on screen.
   public let renderer: Renderer
 
+  private let passes: [FiberReconcilerPass]
```
MaxDesiatov, line 34: nit: needs a doc comment

Sources/TokamakCore/Fiber/FiberReconciler.swift
```diff
@@ -42,27 +46,56 @@ public final class FiberReconciler<Renderer: FiberRenderer> {
...
+      }
     }
+  }
 
-    static func _makeView(_ inputs: ViewInputs<Self>) -> ViewOutputs {
-      .init(
-        inputs: inputs,
-        layoutComputer: { _ in RootLayoutComputer(sceneSize: inputs.content.renderer.sceneSize) }
-      )
+  struct RootLayout: Layout {
```
MaxDesiatov, line 56: nit: doc comment

Sources/TokamakCore/Fiber/Layout/StackLayout.swift
```diff
@@ -0,0 +1,288 @@
...
+}
+
+/// The `Layout.Cache` for `StackLayout` conforming types.
+@_spi(TokamakCore)
+public struct StackLayoutCache {
+  /// The widest/tallest (depending on the `axis`) subview.
+  /// Used to place subviews along the `alignment`.
+  var maxSubview: ViewDimensions?
+  /// The ideal size for each subview as computed in `sizeThatFits`.
+  var idealSizes = [CGSize]()
```
MaxDesiatov, line 37: formatting nit
```suggestion
  /// The widest/tallest (depending on the `axis`) subview.
  /// Used to place subviews along the `alignment`.
  var maxSubview: ViewDimensions?

  /// The ideal size for each subview as computed in `sizeThatFits`.
  var idealSizes = [CGSize]()
```

Sources/TokamakCore/Fiber/Layout/StackLayout.swift
```diff
@@ -0,0 +1,288 @@
...
+
+/// The protocol all built-in stacks conform to.
+/// Provides a shared implementation for stack layout logic.
+@_spi(TokamakCore)
+public protocol StackLayout: Layout where Cache == StackLayoutCache {
+  /// The direction of this stack. `vertical` for `VStack`, `horizontal` for `HStack`.
+  static var orientation: Axis { get }
+  /// The full `Alignment` with an ignored value for the main axis.
+  var _alignment: Alignment { get }
+  var spacing: CGFloat? { get }
```
MaxDesiatov, line 59:
```suggestion
  /// The direction of this stack. `vertical` for `VStack`, `horizontal` for `HStack`.
  static var orientation: Axis { get }

  /// The full `Alignment` with an ignored value for the main axis.
  var _alignment: Alignment { get }
  var spacing: CGFloat? { get }
```

Sources/TokamakCore/Fiber/Layout/StackLayout.swift
```diff
@@ -0,0 +1,288 @@
...
+        min: min.size,
+        max: max,
+        infiniteMainAxis: max[keyPath: Self.mainAxis] == .infinity,
+        spacing: spacing
+      )
+    }
+
+    // Calculate ideal sizes for each View based on their min/max sizes and the space available.
+    var available = proposal[keyPath: Self.mainAxis] - minSize - totalSpacing
+    /// The final resulting size.
```
MaxDesiatov, line 172: nit
```suggestion
    // The final resulting size.
```

Sources/TokamakCore/Fiber/Layout/StackLayout.swift
```diff
@@ -0,0 +1,288 @@
...
+      if $0.view.priority == $1.view.priority {
+        // If the priorities match, allow non-flexible `View`s to size first.
+        return $1.infiniteMainAxis && !$0.infiniteMainAxis
+      } else {
+        return $0.view.priority > $1.view.priority
+      }
+    }) {
+      /// The amount of space available to `View`s with this priority value.
+      let priorityAvailable = available + prioritySize[subview.view.priority, default: 0]
+      /// The number of `View`s with this priority value remaining as a `CGFloat`.
```
MaxDesiatov, line 186: nit
```suggestion
      // The amount of space available to `View`s with this priority value.
      let priorityAvailable = available + prioritySize[subview.view.priority, default: 0]
      // The number of `View`s with this priority value remaining as a `CGFloat`.
```

Sources/TokamakCore/Fiber/Layout/StackLayout.swift
```diff
@@ -0,0 +1,288 @@
...
+    in bounds: CGRect,
+    proposal: ProposedViewSize,
+    subviews: Subviews,
+    cache: inout Cache
+  ) {
+    /// The current progress along the `mainAxis`.
+    var position = CGFloat.zero
+    /// The offset of the `_alignment` in the `maxSubview`,
+    /// used as the reference point for alignments along this axis.
+    let alignmentOffset = cache.maxSubview?[alignment: _alignment, in: Self.orientation] ?? .zero
```
MaxDesiatov, line 218: nit: `///` on local declarations won't have any effect
```suggestion
    // The current progress along the `mainAxis`.
    var position = CGFloat.zero
    // The offset of the `_alignment` in the `maxSubview`,
    // used as the reference point for alignments along this axis.
    let alignmentOffset = cache.maxSubview?[alignment: _alignment, in: Self.orientation] ?? .zero
```

Sources/TokamakCore/Fiber/Layout/ViewSpacing.swift
```diff
@@ -0,0 +1,104 @@
...
+///
+/// When computing spacing in a custom `Layout`, use `distance(to:along:)`
+/// to find the smallest spacing needed to accommodate the preferences
+/// of the `View`s you are aligning.
+public struct ViewSpacing {
+  /// The `View` type this `ViewSpacing` is for.
+  /// Some `View`s prefer different spacing based on the `View` they are adjacent to.
+  @_spi(TokamakCore)
+  public var viewType: Any.Type?
+  private var top: (ViewSpacing) -> CGFloat
```
MaxDesiatov, line 30: formatting readability nit
```suggestion
  /// The `View` type this `ViewSpacing` is for.
  /// Some `View`s prefer different spacing based on the `View` they are adjacent to.
  @_spi(TokamakCore)
  public var viewType: Any.Type?

  private var top: (ViewSpacing) -> CGFloat
```

Sources/TokamakCore/Fiber/Layout/_FrameLayout+Layout.swift
```diff
@@ -0,0 +1,95 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//  Created by Carson Katri on 5/28/22.
+//
+
+import Foundation
+
+private struct FrameLayout: Layout {
```
MaxDesiatov, line 20: nit: doc comment

Sources/TokamakCore/Fiber/Passes/FiberReconcilerPass.swift
```diff
@@ -0,0 +1,137 @@
...
+    struct LayoutCache {
+      /// The erased `Layout.Cache` value.
+      var cache: AnyLayout.Cache
+      /// Cached values for `sizeThatFits` calls.
+      var sizeThatFits: [SizeThatFitsRequest: CGSize]
+      /// Cached values for `dimensions(in:)` calls.
+      var dimensions: [SizeThatFitsRequest: ViewDimensions]
+      /// Does this cache need to be updated before using?
+      /// Set to `true` whenever the subviews or the container changes.
+      var isDirty: Bool
```
MaxDesiatov, line 37:
```suggestion
      /// The erased `Layout.Cache` value.
      var cache: AnyLayout.Cache

      /// Cached values for `sizeThatFits` calls.
      var sizeThatFits: [SizeThatFitsRequest: CGSize]

      /// Cached values for `dimensions(in:)` calls.
      var dimensions: [SizeThatFitsRequest: ViewDimensions]

      /// Does this cache need to be updated before using?
      /// Set to `true` whenever the subviews or the container changes.
      var isDirty: Bool
```

Sources/TokamakCore/Fiber/Passes/ReconcilePass.swift
```diff
@@ -0,0 +1,339 @@
...
+    reconcileRoot: FiberReconciler<R>.Fiber,
+    caches: FiberReconciler<R>.Caches
+  ) where R: FiberRenderer {
+    var node = root
+
+    /// Enabled when we reach the `reconcileRoot`.
+    var shouldReconcile = false
+
+    /// Traits that should be attached to the nearest rendered child.
+    var pendingTraits = _ViewTraitStore()
```
MaxDesiatov, line 75:
```suggestion
    // Enabled when we reach the `reconcileRoot`.
    var shouldReconcile = false

    // Traits that should be attached to the nearest rendered child.
    var pendingTraits = _ViewTraitStore()
```

Sources/TokamakCore/Fiber/Passes/ReconcilePass.swift
```diff
@@ -0,0 +1,339 @@
...
+        if node.fiber?.element == nil {
+          pendingTraits = traits
+        }
+      }
+      // Clear the pending traits once they have been applied to the target.
+      if node.fiber?.element != nil && !pendingTraits.values.isEmpty {
+        pendingTraits = .init()
+      }
+
+      // Ensure the TreeReducer can access any necessary state.
```
MaxDesiatov, line 110: nit
```suggestion
      // Ensure the `TreeReducer` can access any necessary state.
```

Sources/TokamakCore/Fiber/Passes/ReconcilePass.swift
```diff
@@ -0,0 +1,339 @@
...
+          caches.layoutSubviews[parentKey, default: .init(elementParent)].storage.append(subview)
+        }
+      }
+
+      // Setup the alternate if it doesn't exist yet.
+      if node.fiber?.alternate == nil {
+        _ = node.fiber?.createAndBindAlternate?()
+      }
+
+      // Walk all down all the way into the deepest child.
```
MaxDesiatov, line 196:
```suggestion
      // Walk down all the way into the deepest child.
```

Sources/TokamakCore/Fiber/Passes/ReconcilePass.swift
```diff
@@ -0,0 +1,339 @@
+// Copyright 2022 Tokamak contributors
```
MaxDesiatov, line 1: Should this file be named `ReconcilerPass.swift`, not `ReconcilePass.swift`?

Sources/TokamakCore/Fiber/ViewArguments.swift
```diff
@@ -22,6 +22,7 @@ public struct ViewInputs<V> {
   public let content: V
   @_spi(TokamakCore)
   public let environment: EnvironmentBox
+  public let traits: _ViewTraitStore?
```
MaxDesiatov, line 25:
```suggestion
  public let content: V

  @_spi(TokamakCore)
  public let environment: EnvironmentBox

  public let traits: _ViewTraitStore?
```

Sources/TokamakCore/Views/Layout/HStack.swift
```diff
@@ -27,7 +27,8 @@ public let defaultStackSpacing: CGFloat = 8
 ///     }
 public struct HStack<Content>: View where Content: View {
   public let alignment: VerticalAlignment
-  let spacing: CGFloat
+  @_spi(TokamakCore)
+  public let spacing: CGFloat?
```
MaxDesiatov, line 31: formatting readability nit: at a glance this looks like `@_spi(TokamakCore)` is applied to both properties without the newline
```suggestion
  @_spi(TokamakCore)
  public let spacing: CGFloat?

```

MaxDesiatov (commented):
Seems legit, great work!

I only have formatting and doc comment nits.

Sources/TokamakStaticHTML/StaticHTMLFiberRenderer.swift
```diff
@@ -188,16 +207,42 @@ public struct StaticHTMLFiberRenderer: FiberRenderer {
...
+
+  public func render<V: View>(_ view: V) -> String {
+    _ = FiberReconciler(self, view)
+    return """
+    <!doctype html>
+    <html>
+    \(rootElement.description)
+    </html>
+    """
+  }
```
MaxDesiatov, line 247: will the presence of `doctype` here make https://github.com/TokamakUI/Tokamak/pull/489 obsolete?

Package.swift
carson-katri: I think the workflow runs every test target in the package already.

MaxDesiatov: Ah, that's great! Sorry about the confusion.

Sources/TokamakCore/Fiber/Passes/ReconcilePass.swift
carson-katri: The filename matches the struct in this file. Unless you think the struct should also be called `Reconcile**r**Pass` not `ReconcilePass`.

Sources/TokamakStaticHTML/StaticHTMLFiberRenderer.swift
carson-katri: I believe that PR deals with the stack reconciler `StaticHTMLRenderer`, not this fiber reconciler one.
Although this could be enhanced to support meta tags in the head and such once preferences are supported in the fiber reconciler.

Sources/TokamakCore/Fiber/Passes/ReconcilePass.swift
MaxDesiatov: Yeah, I think it's a little bit confusing that it conforms to `FiberReconcilerPass`, but the name of the struct uses verb form, not noun form. I also would expect the name of the struct to be more specific than the name of the struct, but I don't have a good suggestion for a new name right now.

carson-katri:
I made a few changes to (1) move the LayoutSubview logic into the type instead of in the `ReconcilePass`, and (2) remove `layout` values from Views that don't participate in layout.

I was hoping this could help with stack overflows when using dynamic layout. But now I think the only real way to solve this is to make the calls into `LayoutSubview` iterative. I have not been able to figure out how to do this yet. Currently the `FiberReconciler` without dynamic layout enabled can handle ~400 nested views without overflowing, while dynamic layout can only handle ~15 unless you increase the stack size. For now, we can just use recursion though.

MaxDesiatov (approved):
Great, thanks!

PR #500, opened 2022-06-27
Support `Image` in Fiber renderers

carson-katri:
This adds support for the `Image` view, as well as the `aspectRatio` modifier. With dynamic layout enabled, it computes the natural size of the image by loading it and setting an `intrinsicSize` state property on the `Image` to update the layout.

Aspect ratio support was also added to allow for images to be scaled while keeping their resolution.

### DIFF
--- a/Sources/TokamakCore/Views/Image.swift
+++ b/Sources/TokamakCore/Views/Image.swift
@@ -101,10 +101,15 @@ private class ResizableProvider: _AnyImageProviderBox {
 }
 
 public struct Image: _PrimitiveView, Equatable {
-  let provider: _AnyImageProviderBox
+  @_spi(TokamakCore)
+  public let provider: _AnyImageProviderBox
 
   @Environment(\.self)
   var environment
+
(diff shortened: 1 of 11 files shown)


Sources/TokamakCore/Views/Image.swift
```diff
@@ -101,11 +101,16 @@ private class ResizableProvider: _AnyImageProviderBox {
...
-  let provider: _AnyImageProviderBox
+  @_spi(TokamakCore)
+  public let provider: _AnyImageProviderBox
 
   @Environment(\.self)
   var environment
 
+  @_spi(TokamakCore)
+  @State
+  public var intrinsicSize: CGSize?
```
MaxDesiatov, line 112: shouldn't this property be underscored as SPI?

```suggestion
  public var _intrinsicSize: CGSize?
```

MaxDesiatov (commented):
LGTM, just a minor comment for an SPI property.

MaxDesiatov (approved):
Thanks!

PR #501, opened 2022-06-29
Add support for preferences, `@StateObject`, `@EnvironmentObject`, and custom `DynamicProperty` types

carson-katri:
The following are now supported in the Fiber reconciler:
* Preferences
    * To support preferences keys, I made the `ReconcilePass` always start at the root of the view hierarchy (not just when dynamic layout is enabled), but it still only looks for changes once it reaches the first changed fiber.
* `@StateObject`/`@ObservedObject`
* `@EnvironmentObject`
* `@AppStorage`/`@SceneStorage`
* Custom `DynamicProperty` types

> Currently, `backgroundPreferenceValue` does not correctly access the preference values. `overlayPreferenceValue` works because the overlay is evaluated after the underlying content. To fix `backgroundPreferenceValue` we could evaluate the background after the foreground and set the `z-index` to put it underneath.

A new internal API for testing the `FiberReconciler` using `TestFiberRenderer` was also added when creating tests for the new dynamic properties:
```swift
struct TestView: View {
  @State private var count = 0

  var body: some View {
    Button(action: { count += 1 }) {
      Text("\(count)")
        .identified(by: "count") // Use `identified(by:)` to mark views for later retrieval.
    }
    .identified(by: "increment")
  }
}

// Setup a reconciler for `TestView`.
let reconciler = TestFiberRenderer(.root, size: .zero).render(TestView())

// Get a proxy to a View with the id "count".
let count = reconciler.findView(id: "count", as: Text.self)
let increment = reconciler.findView(id: "increment", as: Button<Text>.self)

// Access `.view` to resolve the current content of the fiber.
XCTAssertEqual(count.view, Text("0"))

// Use dynamic member lookup to access properties.
increment.action?()

// Make sure the state changed. `view` is computed when accessed, so it can be used across reconcile passes (despite `current`/`alternate` swaps).
XCTAssertEqual(count.view, Text("1"))
```

### DIFF
--- a/Sources/TokamakCore/Fiber/Fiber+CustomDebugStringConvertible.swift
+++ b/Sources/TokamakCore/Fiber/Fiber+CustomDebugStringConvertible.swift
@@ -22,7 +22,7 @@ extension FiberReconciler.Fiber: CustomDebugStringConvertible {
     {
       return "Text(\"\(text.storage.rawText)\")"
     }
-    return "\(typeInfo?.name ?? "Unknown") \(ObjectIdentifier(self).debugDescription.split(separator: "(")[1].split(separator: ")")[0])"
+    return "\(typeInfo?.name ?? "Unknown")"
   }
 
   private func flush(level: Int = 0) -> String {
--- a/Sources/TokamakCore/Fiber/Fiber+CustomDebugStringConvertible.swift
+++ b/Sources/TokamakCore/Fiber/Fiber+CustomDebugStringConvertible.swift
@@ -22,7 +22,7 @@ extension FiberReconciler.Fiber: CustomDebugStringConvertible {
     {
       return "Text(\"\(text.storage.rawText)\")"
     }
-    return typeInfo?.name ?? "Unknown"
+    return "\(typeInfo?.name ?? "Unknown") \(ObjectIdentifier(self).debugDescription.split(separator: "(")[1].split(separator: ")")[0])"
   }
 
   private func flush(level: Int = 0) -> String {
--- a/Sources/TokamakCore/Fiber/Fiber+CustomDebugStringConvertible.swift
+++ b/Sources/TokamakCore/Fiber/Fiber+CustomDebugStringConvertible.swift
@@ -17,12 +17,13 @@
 
 extension FiberReconciler.Fiber: CustomDebugStringConvertible {
   public var debugDescription: String {
+    let memoryAddress = String(format: "%010p", unsafeBitCast(self, to: Int.self))
     if case let .view(view, _) = content,
        let text = view as? Text
     {
-      return "Text(\"\(text.storage.rawText)\")"
+      return "Text(\"\(text.storage.rawText)\") (\(memoryAddress))"
     }
-    return "\(typeInfo?.name ?? "Unknown")"
+    return "\(typeInfo?.name ?? "Unknown") (\(memoryAddress))"
   }
 
   private func flush(level: Int = 0) -> String {
--- a/Sources/TokamakCore/Fiber/Fiber.swift
+++ b/Sources/TokamakCore/Fiber/Fiber.swift
@@ -94,6 +94,9 @@ public extension FiberReconciler {
     /// Boxes that store `State` data.
     var state: [PropertyInfo: MutableStorage] = [:]
 
+    /// Storage for `PreferenceKey` values as they are passed up the tree.
+    var preferences: _PreferenceStore = .init()
+
     /// The computed dimensions and origin.
     var geometry: ViewGeometry?
 
@@ -175,6 +178,7 @@ public extension FiberReconciler {
         let alternate = Fiber(
           bound: alternateView,
           state: self.state,
+          preferences: self.preferences,
           layout: self.layout,
           alternate: self,
           outputs: self.outputs,
@@ -205,6 +209,7 @@ public extension FiberReconciler {
     init<V: View>(
       bound view: V,
       state: [PropertyInfo: MutableStorage],
+      preferences: _PreferenceStore,
       layout: AnyLayout!,
       alternate: Fiber,
       outputs: ViewOutputs,
@@ -224,6 +229,7 @@ public extension FiberReconciler {
       self.typeInfo = typeInfo
       self.outputs = outputs
       self.state = state
+      self.preferences = preferences
       if element != nil {
         self.layout = layout
       }
@@ -326,6 +332,7 @@ public extension FiberReconciler {
         let alternate = Fiber(
           bound: alternateApp,
           state: self.state,
+          preferences: self.preferences,
           layout: self.layout,
           alternate: self,
           outputs: self.outputs,
@@ -341,6 +348,7 @@ public extension FiberReconciler {
     init<A: App>(
       bound app: A,
       state: [PropertyInfo: MutableStorage],
+      preferences: _PreferenceStore,
       layout: AnyLayout?,
       alternate: Fiber,
       outputs: SceneOutputs,
@@ -358,6 +366,7 @@ public extension FiberReconciler {
       self.typeInfo = typeInfo
       self.outputs = outputs
       self.state = state
+      self.preferences = preferences
       self.layout = layout
       content = content(for: app)
     }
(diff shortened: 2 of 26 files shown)


Sources/TokamakCore/Fiber/Fiber+CustomDebugStringConvertible.swift
```diff
@@ -22,7 +22,7 @@ extension FiberReconciler.Fiber: CustomDebugStringConvertible {
     {
       return "Text(\"\(text.storage.rawText)\")"
     }
-    return typeInfo?.name ?? "Unknown"
+    return "\(typeInfo?.name ?? "Unknown") \(ObjectIdentifier(self).debugDescription.split(separator: "(")[1].split(separator: ")")[0])"
```
MaxDesiatov, line 25: Can we guarantee that `ObjectIdentifier(self).debugDescription.split(separator: "(")` will always return at least two elements? Maybe we need to guard for that?

MaxDesiatov:
Just had a super quick look, will review the rest of it by the end of this week.

Sources/TokamakCore/Fiber/Fiber+CustomDebugStringConvertible.swift
carson-katri: I just removed this from the description, since I was just using it to debug something. If you think I should keep the memory address though, using `unsafeBitCast` may be a better method?

MaxDesiatov: Not unless you found seeing memory addresses particularly useful during development 🙂

Sources/TokamakCore/Fiber/FiberReconciler.swift
```diff
@@ -39,6 +39,7 @@ public final class FiberReconciler<Renderer: FiberRenderer> {
   private let caches: Caches
 
   private var isReconciling = false
+  private var changedFibers = Set<ObjectIdentifier>()
```
MaxDesiatov, line 42: nit: probably worth adding a doc comment explaining why these need to be tracked in a `Set` and they're used throughout the reconciler

Sources/TokamakCore/Fiber/Fiber.swift
```diff
@@ -83,22 +85,33 @@ public extension FiberReconciler {
...
+    /// The nearest parent that receives preferences.
+    unowned var preferenceParent: Fiber?
+
     /// The cached type information for the underlying `View`.
     var typeInfo: TypeInfo?
 
     /// Boxes that store `State` data.
     var state: [PropertyInfo: MutableStorage] = [:]
 
+    /// Subscribed `Cancellable`s
```
MaxDesiatov, line 103: nit: please expand these doc comment to explain why `PropertyInfo` is used as a dictionary key

Sources/TokamakCore/Fiber/FiberReconciler.swift
```diff
@@ -211,18 +206,25 @@ public final class FiberReconciler<Renderer: FiberRenderer> {
...
 
-  func reconcile(from updateRoot: Fiber) {
-    isReconciling = true
-    let root: Fiber
-    if renderer.useDynamicLayout {
-      // We need to re-layout from the top down when using dynamic layout.
-      root = current
-    } else {
-      root = updateRoot
+  func fiberChanged(_ fiber: Fiber) {
```
MaxDesiatov, line 209: this deserves a doc comment too, explaining how it's used

Sources/TokamakCore/Fiber/FiberReconciler.swift
```diff
@@ -211,18 +206,25 @@ public final class FiberReconciler<Renderer: FiberRenderer> {
...
+    let shouldSchedule = changedFibers.isEmpty
+    changedFibers.insert(ObjectIdentifier(alternate))
+    if shouldSchedule {
+      renderer.schedule { [weak self] in
+        self?.reconcile()
+      }
     }
+  }
+
+  func reconcile() {
```
MaxDesiatov, line 221: nit: doc comment

Sources/TokamakCore/Fiber/FiberRenderer.swift
```diff
@@ -57,6 +57,11 @@ public protocol FiberRenderer {
     proposal: ProposedViewSize,
     in environment: EnvironmentValues
   ) -> CGSize
+
+  /// Run `action` on the next runloop.
```
MaxDesiatov, line 61: nit
```suggestion
  /// Run `action` on the next runloop iteration.
```

Sources/TokamakCore/Fiber/FiberRenderer.swift
```diff
@@ -57,6 +57,11 @@ public protocol FiberRenderer {
...
+  ///
+  /// Called by the `FiberReconciler` to perform all reconciliation at once.
```
MaxDesiatov, line 63: could you elaborate in the comment what `at once` means in this context and why performing it all at once might be needed?

Sources/TokamakCore/Fiber/ViewArguments.swift
```diff
@@ -21,10 +21,14 @@ import Foundation
 public struct ViewInputs<V> {
   public let content: V
 
+  public let updateContent: ((inout V) -> ()) -> ()
```
MaxDesiatov, line 24: needs a doc comment

Sources/TokamakCore/Fiber/ViewArguments.swift
```diff
@@ -33,7 +37,11 @@ public struct ViewOutputs {
   /// This is stored as a reference to avoid copying the environment when unnecessary.
   let environment: EnvironmentBox
 
-  let preferences: _PreferenceStore
+  let preferenceStore: _PreferenceStore?
+
+  let transformPreferences: ((_PreferenceStore) -> ())?
+
+  let preferenceAction: ((_PreferenceStore) -> ())?
```
MaxDesiatov, line 44: these two definitely need a doc comment to clarify the difference between them, as they have the same time, which doesn't make it easier to infer their purpose

Sources/TokamakCore/Fiber/walk.swift
```diff
@@ -15,20 +15,23 @@
...
+public enum WalkResult<Renderer: FiberRenderer, Success> {
   case success(Success)
   case finished
   case paused(at: FiberReconciler<Renderer>.Fiber)
 }
 
+@_spi(TokamakCore)
 @discardableResult
-func walk<Renderer: FiberRenderer>(
+public func walk<Renderer: FiberRenderer>(
```
MaxDesiatov, line 34: nit: missing doc comment

Sources/TokamakCore/Preferences/PreferenceKey.swift
```diff
@@ -48,30 +77,69 @@ public extension _PreferenceValue {
...
+    if let existing = values[String(reflecting: key)] {
+      storage = existing
+    } else {
+      storage = .init(key)
+      values[String(reflecting: key)] = storage
+    }
+    return _PreferenceValue(storage: storage)
+  }
+
+  func previousValue<Key>(forKey key: Key.Type = Key.self) -> _PreferenceValue<Key>
```
MaxDesiatov, line 109: nit: missing doc comment

Sources/TokamakTestRenderer/TestViewProxy.swift
```diff
@@ -0,0 +1,102 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//  Created by Carson Katri on 6/30/22.
+//
+
+import Foundation
+@_spi(TokamakCore)
+import TokamakCore
```
MaxDesiatov, line 20: nit: leading newline for `@_spi` looks more readable to me here, but feel free to keep as is at your preference
```suggestion
import Foundation

@_spi(TokamakCore)
import TokamakCore
```

MaxDesiatov (commented):
Overall seems legit, just a few nits with requests for doc comments

MaxDesiatov (approved):
Seems legit, thanks!

PR #502, opened 2022-07-05
Allow window resizing in Fiber renderers

carson-katri:
This converts `sceneSize` to a `CurrentValueSubject` which allows renderers to notify the reconciler when the window changes size.

In the `DOMFiberRenderer`, a `ResizeObserver` watches the `<body>` for size changes.


MaxDesiatov (approved):
Seems legit, thanks!

PR #503, opened 2022-07-06
Support `<meta>`/`<title>` in Fiber renderers

carson-katri:
This adds compatibility with the `.htmlMeta()` and `.htmlTitle()` modifiers added in #483. 2 special cases were also added:

1. `WindowGroup` has a `title` parameter, which can now be used to specify the `<title>` tag in Fiber renderers by extending the `_WindowGroupTitle` view.

2. In SwiftUI, if a `_BackgroundStyleModifier` reaches the top of the scene, it will extend into the top safe area. This PR adds the `"theme-color"` meta tag when the background style reaches the top of the scene:

<table>
<tr>
<td>
  Simulator Screen Shot - iPhone 13 Pro - 2022-07-06 at 10 49 28
</td>
<td>
  Screen Shot 2022-07-06 at 10 42 51 AM
</td>
</tr>
</table>

> While Safari will automatically detect the theme color without the additions from this PR, it will not pick up background color changes due to state. The changes from this PR allow that.

### DIFF
--- a/Sources/TokamakCore/App/Scenes/WindowGroup.swift
+++ b/Sources/TokamakCore/App/Scenes/WindowGroup.swift
@@ -82,7 +82,6 @@ public struct WindowGroup<Content>: Scene, TitledScene where Content: View {
 //  }
 
   public func _visitChildren<V>(_ visitor: V) where V: SceneVisitor {
-    print("Visiting scene")
     visitor.visit(Group {
       _WindowGroupTitle(title: self.title)
       content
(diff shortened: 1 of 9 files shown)


Sources/TokamakCore/App/Scenes/WindowGroup.swift
```diff
@@ -15,6 +15,11 @@
 //  Created by Carson Katri on 7/16/20.
 //
 
+@_spi(TokamakCore)
+public struct _WindowGroupTitle: _PrimitiveView {
+  public let title: Text?
```
ezraberch, line 20: Does this need to be an optional?

In `WindowGroup`s where `title` is nil, it should be more efficient to not use a `_WindowGroupTitle` at all.

PR #506, opened 2022-07-16
add missing length unit to width and height attributes

astrotuna201:
PR fixes missing "px" units to width and height attribute, that resulted in attribute being dropped as invalid by browsers.


MaxDesiatov (changes requested):
Would you be able to add a test (possibly a snapshot test if you find that more appropriate) to prevent regressions in the future? Thanks!

PR #510, opened 2022-08-20
Update default assignee in `bug_report.md`

MaxDesiatov:
As I'm going to step away as a maintainer, I'd like to avoid being assigned for new issues by default.

Let me know if you'd like the assignee field to be empty by default.


kateinoigakukun (approved):

carson-katri:
This is ok with me. Thank you for getting this project started!

PR #515, opened 2022-09-06
Change (V/H)Stack implementation from grid to flex

Tyler-pollick:
Completed @SwiftCoderJoe conversion of (V/H)Stack from grid to flex (PR [#466](https://github.com/TokamakUI/Tokamak/pull/466)).

As @SwiftCoderJoe mentioned in his PR, this change should fix the Spacer issues described in #441.

All relevant tests have been updated. I will circle back around and back fill with more (V/H)Stack tests.


SwiftCoderJoe:
It's interesting to me that this works with the ZStack... I thought I had applied the same changes from grid syntax to flex syntax and then added minimum heights and widths to 0, but either I for some reason never tested it with this configuration or you have something special that I missed! Either way, I'm really glad to see this finally completed and I'm glad we're moving forward on this! Thanks!

PR #516, opened 2022-09-18
Enable async/await in DOM Fiber renderer

ahti:
I noticed that Tasks weren't running when using the Fiber renderer. I'm not sure this is the appropriate place, but the normal DOM renderer calls `installGlobalExecutor` in it's `init`, so I just mirrored that.


carson-katri (approved):
This looks good to me, thanks!

PR #517, opened 2022-09-23
[DNM] Check swift-wasm-5.7-SNAPSHOT-2022-09-18-a

kateinoigakukun:


kateinoigakukun:
OK, I confirmed the latest snapshot works well :)

PR #525, opened 2022-10-31
Fiber Reconciler fixes

### LINKED ISSUE #523: Fiber layout/reconciliation issues
<details><summary>Sample code</summary><p>

```
import TokamakDOM
import Foundation

@main
struct TokamakApp: App {
    static let _configuration: _AppConfiguration = .init(reconciler: .fiber(useDynamicLayout: false))
    var body: some Scene { WindowGroup("Tokamak App") { ContentView() } }
}

enum State {
    case a
    case b([String])
    case c(String, [Int])
    case d(String, [Int], String)
}

final class StateManager: ObservableObject {
    private init() { }
    static let shared = StateManager()

    @Published var state = State.a
}

struct ContentView: View {
    @ObservedObject var sm = StateManager.shared

    var body: some View {
        switch sm.state {
        case .a:
            // VStack {
                Button("go to b") {
                    sm.state = .b(["eins", "zwei", "drei"])
                }
            // }
        case .b(let arr):
            VStack {
                Text("b:")
                ForEach(arr, id: \.self) { s in
                    Button(s) {
                        sm.state = .c(s, s == "zwei" ? [1, 2] : [1])
                    }
                }
            }
        case .c(let str, let ints):
            VStack {
                Text("c \(str)")
                    .font(.headline)
                Text("hello there")
                ForEach(ints, id: \.self) { i in
                    let d = "i = \(i)"
                    Button(d) {
                        sm.state = .d(str, ints, d)
                    }
                }
            }
        case .d(_, let ints, let other):
            VStack {
                Text("\(other)")
                Text("count \(ints.count)")
            }
        }
    }
}
```

</p></details>

**Describe the bug**
Using the fiber reconciler, the sample code above misbehaves in various ways:

**To Reproduce**
1. Run the sample code
2. Click "go to b"
3. => The correct buttons for state b appear, but the text does not
4. Click any button
5. => The "i = number" buttons for state c appear, but some from state b are left over, still no text

Changing the configuration to use dynamic layout puts all the buttons on top of each other.

If you add the VStack to wrap the button for state a, and apply the change from #522 the buttons update properly, but the text is still missing.

**Expected behavior**
The way the code behaves when commenting out the `_configuration`. No lingering buttons, all texts show up.

**Desktop:**
 - macOS 12.6
 - Safari 16
 - Version of Tokamak: main branch

ahti:
This is what I have done so far to make the example (with a little modification) from #523 work.

Currently, things work until returning to the initial state, after which pressing the button will leave an empty screen.

Some comments in no particular order:

* The reproducer target is of course only for debugging and to be removed before merge.
* I've left the logs as normal prints for now, since [swift-log doesn't build on wasm rn](https://github.com/apple/swift-log/pull/236), we can remove those as well if things get ready before swift-log is fixed.
* Pay no attention to the individual commits, I'll clean those up once things look more mergeable.
* The implementation of `ReconcilePass.reconcile(_: , in:, caches:)` obviously still needs to be refactored.
* The reconciliation pass currently emits mutations that seem to assume replacing an element transfers any children to the replacement. I've adjusted the DOM and test renderers, but maybe the emitted mutations should be changed instead.
* I tried turning the reproducer into a test case, but that fails one step earlier, while returning to the initial state, due to an out of range index in an insert mutation.
* Another issue I observed: There are cases where a `remove` mutation is emitted with the wrong `parent`. The `DOMRenderer` doesn't use the parent, so it's not an issue there, but it does materialize with the test renderer.
* The parent in the remove mutation is optional, but all places where such a value is created pass a non-nil parent. Is there a reason for this, or can this be turned non-optional?


carson-katri:
I don't recall specifically, but the `parent` may be be optional because the root of the hierarchy has no parent and could hypothetically be removed.

carson-katri:
Tested on a small project and didn't notice any regressions.

ahti:
I finally found some time and started refactoring `ReconcilePass.reconcile` (not quite done with that), and also fixed the empty screen when going through all the states of the reproducer and then from `.a` to `.b` again (nothing was inserted when a fiber that should have an element had an alternate, but the alternate didn't have an element).

There is still an issue with label content not being updated tho, which I'm going to look into next.

ahti:
Fixed the stale label content as well.

The only issue left that I am aware of is the `.remove` mutation with wrong parent. I've extended the test to reproduce that one, too.

I also changed the test renderer so that multiple changed fibers get batched like they do in real renderers. Without that, the test was kind of flaky and would randomly succeed or fail (before adding the part to reproduce the remove bug, that is).

Package.swift
```diff
@@ -15,6 +15,7 @@ let package = Package(
       name: "TokamakDemo",
       targets: ["TokamakDemo"]
     ),
+    .executable(name: "RecRep", targets: ["RecRep"]),
```
carson-katri, line 18: I'm not sure we should have this executable as part of the Tokamak package. It seems you have already adapted it to a test case, which should be enough.

Sources/TokamakCore/Fiber/FiberReconciler.swift
```diff
@@ -270,6 +270,14 @@ public final class FiberReconciler<Renderer: FiberRenderer> {
     self.alternate = current
     current = alternate
 
+    print("""
+      reconcile done
+      mutations were:
+      \(visitor.mutations.map { "  \($0)" }.joined(separator: "\n"))
+      alternate is \(self.alternate.recursiveDescription)
+      current is \(current.recursiveDescription)
+      """)
```
carson-katri, line 279: Can you clean up any print statements before merging? Better debug logging could be tackled in a separate PR.

Sources/TokamakCore/Fiber/Passes/ReconcilePass.swift
```diff
@@ -223,41 +231,73 @@ struct ReconcilePass: FiberReconcilerPass {
...
-          )
+    guard let fiber = node.fiber else { return nil }
+
+    func canUpdate(_ fiber: FiberReconciler<R>.Fiber) -> Bool {
+      fiber.typeInfo?.type == fiber.alternate?.typeInfo?.type
+    }
+
+    invalidateCache(for: fiber, in: reconciler, caches: caches)
+
+    switch (fiber.element, node.newContent, fiber.alternate?.element) {
```
carson-katri, line 242: Very nice improvement here!

Sources/TokamakCore/Fiber/Passes/ReconcilePass.swift
```diff
@@ -223,41 +231,73 @@ struct ReconcilePass: FiberReconcilerPass {
...
+
+    case let (_, nil, alt?):
+      guard let parent = fiber.alternate?.elementParent?.element else { break } // todo: name => altParent?
+
+      fiber.element = nil
+      fiber.elementIndex = 0
+      if let p = fiber.elementParent { caches.elementIndices[ObjectIdentifier(p)]? -= 1 }
+      return .remove(element: alt, parent: parent)
+
+    case let (element?, content, _) where fiber.alternate?.element == nil:
```
carson-katri, line 284: Couldn't this be written as:
```suggestion
    case let (element?, content, nil):
```

carson-katri (changes requested):
Very nice improvements, thank you. I used it on a test project of mine and it worked well.

ahti:
superseeded by #542

PR #526, opened 2022-11-05
Implement Path.addArc(tangent1End:...).

filip-sakel:
The `addArc(tangent1End:tangent2End:radius:transform:)` method was left unimplemented and added nothing to the path components. This implementation uses some vector arithmetic on `CGPoint` to create an arc. There are probably better approaches out there, but I think this a good starting point.

### DIFF
--- a/Sources/TokamakDOM/DOMFiberRenderer.swift
+++ b/Sources/TokamakDOM/DOMFiberRenderer.swift
@@ -310,7 +310,7 @@ public struct DOMFiberRenderer: FiberRenderer {
           fatalError("The previous element does not exist (trying to replace element).")
         }
         let replacementElement = createElement(replacement)
-        _ = parentElement.replaceChild?(previousElement, replacementElement)
+        _ = parentElement.replaceChild?(replacementElement, previousElement)
       case let .update(previous, newContent, geometry):
         previous.update(with: newContent)
         guard let previousElement = previous.reference


filip-sakel:
Swiftlint is complaining about using `a = a + b`, but there’s no reason to implement `+=` just for these operations. I don't know how I can disable this check for the places where this is used.

Sources/TokamakCore/Shapes/Path/PathMutations.swift
```diff
@@ -207,3 +216,111 @@ func getArc(
...
+    sqrt(x * x + y * y)
+  }
+
+  /// Changes the magnitude of the vector on the origin's
+  /// coordinate system.
+  func withMagnitude(_ newValue: Double, origin: CGPoint = .zero) -> CGPoint {
+    var copy = self - origin
+
+    let scale = newValue / copy.magnitude
+    copy = copy * scale
```
carson-katri, line 247:
```suggestion
// swiftlint:disable:next shorthand_operator
    copy = copy * scale
```

carson-katri (changes requested):
Could you add a snapshot test?

filip-sakel:
I saw how other snapshot tests are written and think I understand how to write one, but I'm not sure how I can generate a snapshot to compare against. 

I'll add the snapshot tests and SwiftLint comments later in the week.

carson-katri:
I think you could just run the test and copy the image it writes in, then compare against that going forward. Layout tests compare directly against a rendered SwiftUI view.

filip-sakel:
I realized that my code doesn't really choose the start-end/clockwise combinations correctly. This means that the start of the arc may be on the second tangent and the arc's end will be on the first tangent. 

So, I'll keep working on this to get everything exactly like the native implementations.

filip-sakel:
I fixed the issue and added some more tests. However, I was unable to get the project to build so I could run the snapshot tests. I run the following according to the contribution guide to no avail:

```zsh
swift build --product TokamakPackageTests && `xcrun --find xctest` .build/debug/TokamakPackageTests.xctest
```

I’m not sure why, but there seems to be a problem with core-graphics types, such as `CGLineJoin` and `CGPoint`. @carson-katri could you add the snapshot tests?

Tests/TokamakStaticHTMLTests/RenderingTests/ShapeTests.swift
```diff
@@ -65,6 +65,123 @@ final class ShapeRenderingTests: XCTestCase {
       timeout: defaultSnapshotTimeout
     )
   }
+
+  func testTangentArc() {
```
carson-katri, line 69: When running this test, the call stack fills in `getArc` where it gets the second part of the chunk: https://github.com/TokamakUI/Tokamak/blob/512933bcbf7232a023a9d40e72801bb1a830602a/Sources/TokamakCore/Shapes/Path/PathMutations.swift#L189
It seems to make sense, as the `chunk1` angle is created by adding `.pi / 2` to the startAngle, which the call to `getArc` then checks and sees the value is greater than `.pi / 2`.

Personally I'm not sure what the solution to this is, I'm hoping you have some ideas 🙃

carson-katri (changes requested):

PR #537, opened 2023-07-19
Fix isPressed for ButtonConfiguration

### LINKED ISSUE #534: ButtonStyle configuration.isPressed not updated properly
**Describe the bug**
ButtonStyle configuration.isPressed not updated properly

**To Reproduce**
The demo code is enough to reproduce the bug
```swift
Button("Pressed Button Style") {
    print("tapped")
}
.buttonStyle(
    PressedButtonStyle(pressedColor: Color.red)
)
```
``` swift
struct PressedButtonStyle: ButtonStyle {
  let pressedColor: Color

  func makeBody(configuration: Configuration) -> some View {
    configuration.label
      .foregroundColor(configuration.isPressed ? pressedColor : .blue)
      .padding(15)
  }
}
```
1. Press button and keep it pressed
2. Move away from the button while still pressing the button
3. Release
4. Observe that `.isPressed` is still set to `true`

**Expected behavior**
`.isPressed` set to `false` and color change

**Desktop (please complete the following information):**
 - OS: [macOS 13.3.1]
 - Browser [chrome, safari]
 - Version of the browser [112.0.5615.137 and 16.4]
 - Version of Tokamak [0.11.0]

ffried:
This should fix #534.

### DIFF
--- a/Sources/TokamakCore/Views/Controls/Button.swift
+++ b/Sources/TokamakCore/Views/Controls/Button.swift
@@ -103,7 +103,7 @@ public struct _Button<Label>: View where Label: View {
   public let action: () -> ()
 
   @State
-  public var isPressed = false
+  public var isPressed: (down: Bool, inside: Bool) = (false, false)
 
   let anyStyle: AnyButtonStyle
   public var style: Any.Type { anyStyle.type }
@@ -112,7 +112,7 @@ public struct _Button<Label>: View where Label: View {
       configuration: .init(
         role: role,
         label: .init(body: AnyView(label)),
-        isPressed: isPressed
+        isPressed: isPressed.down && isPressed.inside
       )
     )
   }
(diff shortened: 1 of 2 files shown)


Sources/TokamakCore/Views/Controls/Button.swift
```diff
@@ -103,7 +103,7 @@ public struct _Button<Label>: View where Label: View {
   public let action: () -> ()
 
   @State
-  public var isPressed = false
+  public var isPressed: (down: Bool, inside: Bool) = (false, false)
```
ffried, line 106: I'm not sure about the public api here. It's a breaking change in theory, but AFAIK `_Button` should only be used internally.

ffried:
@carson-katri I've fixed a small bug in my first implementation (re-entering the button while holding the mouse down didn't work correctly).
Also, the two workflow runs that previously failed don't seem to be related to my changes.

carson-katri (approved):
Great, thanks!

PR #551, opened 2023-09-15
Update readme for Linux installation details

shial4:
After this PR is merged for carton 
https://github.com/swiftwasm/carton/pull/408


shial4:
@carson-katri

